PR #3, opened 2020-05-30
Enable hmac-sha2-512 by fixing javax.crypto.ShortBufferException via …

norrisjeremy:
…doubling the buffer size in certain locations.


norrisjeremy:
FYI a better fix for this would probably be to dynamically reallocate the buffer.
I might take a stab at that later, but I want to try and implement some other items first, such as ETM & possibly Curve25519/Ed25519.

norrisjeremy:
See #4.

PR #4, opened 2020-06-01
Implement more algorithms

norrisjeremy:
1) Implement the following algorithms:
- diffie-hellman-group14-sha256
- diffie-hellman-group16-sha512
- diffie-hellman-group18-sha512
- hmac-md5-etm@openssh.com
- hmac-md5-96-etm@openssh.com
- hmac-sha1-etm@openssh.com
- hmac-sha1-96-etm@openssh.com
- hmac-sha2-256-etm@openssh.com
- hmac-sha2-512-etm@openssh.com
- aes128-gcm@openssh.com
- aes256-gcm@openssh.com
- curve25519-sha256
- curve25519-sha256@libssh.org

2) Fix and enable hmac-sha2-512.

3) Add runtime checks for rsa-sha2-256 & rsa-sha2-512 signature types.

4) Add runtime checks for MACs.

5) Change default algorithms to better match modern OpenSSH.

6) Add integrations tests (disabled by default) to allow testing against OpenSSH.

7) Adapt diffie-hellman-group-exchange to allow for a max size of 8192.

Note:
The support for curve25519-sha256 & curve25519-sha256@libssh.org will only work on Java11+, as it uses features added in JEP 324.
Maven has been adjusted to require Java11+ at buildtime, but still targeting Java8+ at runtime, using a multi-release jar to mask the classes requiring Java11+ at runtime.


norrisjeremy:
FYI, I looked into implementing chacha20-poly1305@openssh.com using features added in JEP 329 in Java11, but it doesn't appear that is possible with JEP 329 alone due to the way in which OpenSSH performs the encryption+mac sequences.

norrisjeremy:
Also, it shouldn't be difficult to add support for curve448-sha512.
But since it isn't supported in OpenSSH, I don't have any way to test it, so I decided to leave it out for now.

mwiede:
Thanks for your contribution. I really like your testing approach!!

norrisjeremy:
Thanks! I needed to find a way to make sure all these changes worked correctly and this seemed like a decent way to do so.

There are some other things I spotted as I was reviewing alot the code that I may circle back to as time permits.

For example, adding more logging in places to allow for better troubleshooting. And there are places that are simply doing a `System.err.println(...)`, which should probably be converted into log statements so that users can capture them easier.

There's also alot of duplicated code in many of the Cipher classes in the `com.jcraft.jsch.jce` package that would probably be refactored. I'm not sure how far you want to diverge from the original upstream?

norrisjeremy:
Ok, sounds good.
FYI, I'm working on `FingerprintHash` support to allow outputting them in either MD5 or SHA256 for more compatibility with OpenSSH.

PR #5, opened 2020-06-03
Expand algorithm tests

norrisjeremy:
1) Add tests for:
- ssh-dss
- blowfish-cbc
- arcfour
- arcfour256
- arcfour128
- zlib@openssh.com
- none

2) Add support for FingerprintHash.

3) Make sure to account for AEAD tag size when checking to resize internal buffer.

4) Make MAC comparisons timing safe.

Note: this update switches the Alpine Linux test container from 3.12 to 3.7, which includes OpenSSH version 7.5, which was the last release that included the blowfish & arcfour ciphers.


norrisjeremy:
I think the build check above may be a transient error?

PR #14, opened 2020-09-29
improve log messages for DHEC key exchange

danielgrahl:
The log message upon receiving SSH_MSG_UNIMPLEMENTED after a DHEC key exchange is a bit confusing. Many users think that the exchange has failed. These changes should make the process of kex more clear.

### DIFF
--- a/src/main/java/com/jcraft/jsch/DHECN.java
+++ b/src/main/java/com/jcraft/jsch/DHECN.java
@@ -52,7 +52,8 @@ public abstract class DHECN extends KeyExchange{
   protected String sha_name; 
   protected int key_size;
 
-  public void init(Session session,
+  @Override
+public void init(Session session,
 		   byte[] V_S, byte[] V_C, byte[] I_S, byte[] I_C) throws Exception{
     this.session=session;
     this.V_S=V_S;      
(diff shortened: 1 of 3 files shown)


src/main/java/com/jcraft/jsch/DHECN.java
```diff
@@ -109,6 +108,7 @@ public boolean next(Buffer _buf) throws Exception{
     int i,j;
     switch(state){
     case SSH_MSG_KEX_ECDH_REPLY:
+                JSch.getLogger().log( Logger.INFO, "Received SSH_MSG_KEX_ECDH_REPLY" );
```
norrisjeremy, line 111: I think this should be conditioned on `if(JSch.getLogger().isEnabled(Logger.INFO)){`.

src/main/java/com/jcraft/jsch/DHECN.java
```diff
@@ -96,8 +96,7 @@ public void init(Session session,
     session.write(packet);
 
     if(JSch.getLogger().isEnabled(Logger.INFO)){
-      JSch.getLogger().log(Logger.INFO, 
-                           "SSH_MSG_KEX_ECDH_INIT sent");
+            JSch.getLogger().log( Logger.INFO, "SSH_MSG_KEX_ECDH_INIT sent with EC nistp" + key_size );
```
norrisjeremy, line 99: Do we want to amend DHXEC in a similar manner to provide and indication which curve was sent?

norrisjeremy: In fact, I'm wondering if it might be best to simply tweak all the DH* classes to just include the kex name?

danielgrahl: My experience of working with a different servers: often in their EC implementations something is broken; it helps a lot finding connection errors when you know on what curve we did agree.

PR #15, opened 2020-10-12
Fix bug with AEAD ciphers when compression is used.

norrisjeremy:
The inflater code expects the end of buffer to not include the AEAD tag.

I updated the integration tests to always test every cipher & mac both with and without compression enabled to hopefully catch issues like this in the future.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -193,6 +193,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-javadoc-plugin</artifactId>
+                <version>3.2.0</version>
                 <configuration>
                     <detectJavaApiLink>false</detectJavaApiLink>
                 </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -128,11 +128,17 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <version>2.22.2</version>
+                <configuration>
+                    <trimStackTrace>false</trimStackTrace>
+                </configuration>
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-failsafe-plugin</artifactId>
                 <version>2.22.2</version>
+                <configuration>
+                    <trimStackTrace>false</trimStackTrace>
+                </configuration>
                 <executions>
                     <execution>
                         <goals>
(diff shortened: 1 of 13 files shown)


norrisjeremy:
I'm not sure what caused this failure with just ea4283a applied earlier:
```
Error:  Tests run: 76, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 116.501 s <<< FAILURE! - in com.jcraft.jsch.AlgorithmsIT
Error:  testCiphers{String, String}[7]  Time elapsed: 1.061 s  <<< ERROR!
com.jcraft.jsch.JSchException: Session.connect: java.io.IOException: End of IO Stream Read
	at com.jcraft.jsch.AlgorithmsIT.doSftp(AlgorithmsIT.java:391)
	at com.jcraft.jsch.AlgorithmsIT.testCiphers(AlgorithmsIT.java:244)
```

It seems to have gone away after I pushed 1393e81 to try and get more details...

mwiede:
I also had test failures with my latest changes. Some flakiness, because when re-running the maven job without any code changes, it went through. Now it failed again after re-running. We need to investigate.

norrisjeremy:
Yes, I've reproduced the intermittent failures locally and I'm attempting to track them down.
I believe that there is an issue in curve25519-sha2 that may be the cause.

mwiede:
Yes could be. Locally I have OpenJdk 14 and AlgorithmsIT#testJava11KEXs is
failing.

norrisjeremy:
Ok, I apologize for the overall size of this PR now: but I think I've finally tracked down the source of the intermittent test failures.

It appears that the `curve25519-sha2` code I originally submitted earlier this year wasn't correctly handling encoding of the Q_C values.

I've run the tests in a continuous loop locally for a couple hours this afternoon and didn't see any failures.
I'm also continuing to run them in a non-stop loop for another few hours this evening to see how well things turn out.

norrisjeremy:
Ok, I let the tests run in a loop for almost 3 hours locally ( `mvn verify -DskipITs=false` 100 times).
There were only 4 failures, none of which appeared to be related to any crypto/SSH protocol issues:
 - Two instances of `org.testcontainers.containers.ContainerLaunchException: Container startup failed`.
 - One instance of `com.jcraft.jsch.JSchException: channel is not opened`.
 - One instance of `com.jcraft.jsch.JSchException: Session.connect: java.net.SocketTimeoutException: Read timed out`.

I don't believe any of these are caused by latent bugs in the new crypto code, so I believe the issues with the AES-GCM & X25519 implementations may finally be solved with the changes contained in this PR.

norrisjeremy:
FYI, on a separate note: I now have an implementation of `ssh-ed25519` that can be added.
I'll submit a new PR after we get this PR wrapped up.

PR #17, opened 2020-10-15
Add support for chacha20-poly1305@openssh.com, ssh-ed25519, ssh-ed448, curve448-sha512, diffie-hellman-group15-sha512 & diffie-hellman-group17-sha512

norrisjeremy:
This makes use of the new EdDSA feature added in [Java 15's JEP 339](https://openjdk.java.net/jeps/339).
I also tweaked the Github workflow to hopefully:
1. Run tests against Java 8, 11 & 15.
2. Cache local maven repository to speed up builds a bit.
3. Archive entire set of Surefire, Failsafe & JaCoCo reports.

I ran tests locally for a couple hours last night and didn't see any failures, so hopefully there aren't any bugs in the new EdDSA code.

PR #19, opened 2020-10-16
added integration test for public key authentication

mwiede:


src/test/java/com/jcraft/jsch/KeyPairTest.java
```diff
@@ -56,21 +55,20 @@ void loadKey(String path, String password, String publicKeyType) throws URISynta
...
+        });
     }
 
     @Test
     void genKeypair() {
         final JSch jSch = new JSch();
         assertDoesNotThrow(() -> {
             KeyPair kpair = KeyPair.genKeyPair(jSch, KeyPair.RSA, 1024);
-            kpair.writePrivateKey("my-private-key");
+            kpair.writePrivateKey(System.getProperty("java.io.tmpdir") + File.separator + "my-private-key");
```
norrisjeremy, line 71: In #17 I changed this to use a JUnit @Tempdir, which I think would be preferable, that way it gets cleaned up automagically by JUnit framework.

mwiede: Thanks, I did not know this annotation yet :thumbsup: It's definitively better.

Issue #13, opened 2020-07-23
reject HostKey with some servers


artjomsk:
Hello!

The issue is starting from 0.1.58

Old hostkeys (ssh-rsa) are not working. Getting the following error in log:
```
JSCHLogger] (default task-5) kex: server: curve25519-sha256,curve25519-sha256@libssh.org,ecdh-sha2-nistp256,ecdh-sha2-nistp384,ecdh-sha2-nistp521,diffie-hellman-group-exchange-sha256,diffie-hellman-group16-sha512,diffie-hellman-group18-sha512,diffie-hellman-group-exchange-sha1,diffie-hellman-group14-sha256,diffie-hellman-group14-sha1,diffie-hellman-group1-sha1
JSCHLogger] (default task-5) kex: server: ssh-rsa,rsa-sha2-512,rsa-sha2-256,ecdsa-sha2-nistp256,ssh-ed25519
JSCHLogger] (default task-5) kex: server: chacha20-poly1305@openssh.com,aes128-ctr,aes192-ctr,aes256-ctr,aes128-gcm@openssh.com,aes256-gcm@openssh.com,aes128-cbc,aes192-cbc,aes256-cbc,blowfish-cbc,cast128-cbc,3des-cbc
JSCHLogger] (default task-5) kex: server: chacha20-poly1305@openssh.com,aes128-ctr,aes192-ctr,aes256-ctr,aes128-gcm@openssh.com,aes256-gcm@openssh.com,aes128-cbc,aes192-cbc,aes256-cbc,blowfish-cbc,cast128-cbc,3des-cbc
JSCHLogger] (default task-5) kex: server: umac-64-etm@openssh.com,umac-128-etm@openssh.com,hmac-sha2-256-etm@openssh.com,hmac-sha2-512-etm@openssh.com,hmac-sha1-etm@openssh.com,umac-64@openssh.com,umac-128@openssh.com,hmac-sha2-256,hmac-sha2-512,hmac-sha1
JSCHLogger] (default task-5) kex: server: umac-64-etm@openssh.com,umac-128-etm@openssh.com,hmac-sha2-256-etm@openssh.com,hmac-sha2-512-etm@openssh.com,hmac-sha1-etm@openssh.com,umac-64@openssh.com,umac-128@openssh.com,hmac-sha2-256,hmac-sha2-512,hmac-sha1
JSCHLogger] (default task-5) kex: server: none,zlib@openssh.com
[...]

PR #22, opened 2020-10-21
Fix for #13

norrisjeremy:
Fix the problem reported in #13 by reordering the default server_host…_key setting to prefer RSA type keys.
This restores backwards compatibility with releases prior to 0.1.58.


norrisjeremy:
FYI OpenSSH has some logic where by it will examine the known_hosts file to determine any host keys it already has for the server and then reorder the host key algorithms it proposes to prefer the ones it already knows.
I may look into attempting to duplicate this logic into JSch in the future.
But for now, this should restore compatibility with releases prior to 0.1.58.

norrisjeremy:
Ok, I added logic to examine the known host keys and dynamically reorder the server_host_key proposal to prefer types that are already known for the host.

norrisjeremy:
I think I know why there was a CI failure. There is a race condition with the ListAppender class from Logback in which the docker containers continue to output log messages while the unit tests are checking the logs.

I can come up with a fix for that later today or tomorrow and submit it in a new PR.

norrisjeremy:
See #24 for what I believe is a fix for the test failure I noted above.

norrisjeremy:
Hi, just wanted to check on getting this PR merged?

norrisjeremy:
Hi, just wanted to check again on getting this PR merged?

norrisjeremy:
Hi just wanted to check in on getting this merged?

PR #30, opened 2021-02-16
Add jBCrypt tests. Resolves #20.

### LINKED ISSUE #20: Include TestBCrypt.java unit test?
Do you think a copy of the TestBCrypt.java unit tests should also be incorporated into the build?

norrisjeremy:


src/test/java/com/jcraft/jsch/AlgorithmsIT.java
```diff
@@ -474,7 +474,7 @@ private void doSftp(Session session, boolean debugException) throws Exception {
       session.disconnect();
       jschAppender.stop();
       // Sleep for a short bit to allow logs from test container to catch up
-      Thread.sleep(100L);
+      Thread.sleep(250L);
```
norrisjeremy, line 477: The other way to solve this would be to remove the assertions checking the OpenSSH logs from the Docker container.

It seems that the logs from the Docker container are async and may be fully posted by the time the tests start checking them.

PR #55, opened 2021-07-29
Fix bug with partial initialization of a decrypted private key in open SSH format

pavelsher:
Attempt to decrypt an RSA key stored in an encrypted form in an OpenSSH format and then fetch private key throws an exception:
Exception in thread "main" java.lang.NullPointerException
	at com.jcraft.jsch.KeyPairRSA.getPrivateKey(KeyPairRSA.java:105)

This happens because ep_array and eq_array are not initialized when KeyPairRSA is created.


norrisjeremy:
I would have to defer to @mwiede as I'm not as familiar with the various ssh key file formats.

Issue #59, opened 2021-08-09
Flag to enable "legacy" algos on initialization?


Hr0bar:
Hi, I see a lot of users are expecing full in-place compatibility with old Jsch, but that does not happen due to the retirement of some now insecure algos:

https://github.com/mwiede/jsch/issues/48
https://github.com/mwiede/jsch/issues/45
https://github.com/mwiede/jsch/issues/40
https://github.com/mwiede/jsch/issues/37

I propose adding a flag that would enable all algos that were previously enabled in the legacy Jsch, something like:

new com.jcraft.jsch.JSch(enable_insecure = "true");

Of course, they would be the least preferred/offered option in the negotiation list.

Reason:
we use Jsch to connect to all kinds of versions (some 10+ years old) of below OSes, and its simply impossible to live without these legacy algos. I know we can enable them via configuration as in the example existing issues, but a simple flag that would enable all of them at once is definitely a cleaner way.

- 3Com
- A10 ACOS
- Allot
- Arista EOS
- Aruba AOS
- BlueCat
- BlueCoat CAS
- BlueCoat PacketShaper
- BlueCoat ProxyAV
- BlueCoat ProxySG
- Brocade Vyatta NetworkOS
- CheckPoint GAIA
- CheckPoint IPSO
- CheckPoint SPLAT
- Cisco ACE
- Cisco ACS
- Cisco APIC
- Cisco ASA
- Cisco AsyncOS
- Cisco CatOS
- Cisco CSS
- Cisco Express
- Cisco FTD
- Cisco FWSM
- Cisco FXOS
- Cisco IOS
- Cisco IOS XR
- Cisco IPS
- Cisco ISE
- Cisco NXOS
- Cisco Prime
- Cisco WAAS
- Cisco WLC
- Citrix NetScaler
- Dell iDRAC
- Dell NetworkingOS
- Enterasys
- F5
- Forcepoint Stonesoft
- Fortinet FortiOS
- Gigamon GigaVUEOS
- HP
- Huawei
- IBM AlteonOS/BladeOS/NetworkingOS
- IBM ISCLI
- Infoblox NIOS
- Juniper JUNOS
- Juniper ScreenOS
- Lenovo CNOS
- Linux
- Nokia VitalQIP
- Nortel WSS
- Oracle Acme PacketOS
- Palo Alto PANOS
- PulseSecure
- Radware Alteon
- Riverbed RiOS
- Spectracom
- TrendMicro IPS TOS
[...]

Issue #47, opened 2021-05-19
Authentication with public key fails with 0.1.62


avnerw74:
Hi,

We were using 0.1.55 to connect to a server using public key auth without a problem.
After we upgraded to 0.1.62, public key authentication fails with "Auth Fail".
If we use the UserAuthPublicKey class from 0.1.55, we are able to login again. 

Is this a known bug?

Thanks,
Avner

mwiede:
Hi Avner,
I guess that you have the same issue like the guys in #37 or #40

The reason is, that configured algorithms is changed in 0.1.62 for security reasons. They are aligned with recent openssh versions.

So you might have to change the config like
`Session session = Jsch.getSession(username, host) session.setConfig("kex", "<yourkeyalgo>");` or kex
or
`session.setConfig("cipher.c2s", session.getConfig("cipher.c2s") + ",aes128-cbc");
session.setConfig("cipher.s2c", session.getConfig("cipher.s2c") + ",aes128-cbc");`
for ciphers.

I will update the readme soon, since it is important to understand this change when using this library as a replacement of Jsch.

Kind regards
Matthias

ihudedi:
Hi @mwiede 
We do change the ciphers and in kex  as well as you wrote above but the authentication failed when using public key.
When using password it works fine.
If I use the UserAuthPublicKey  file form version 1.55 the public key authentication succeeded OK.
Thanks,
Itay

norrisjeremy:
Hi Avner & Itay,

A few questions:

1. Are you able to add a JSch logger and capture debug logging from when it fails with 0.1.62?
2. Is the public key you are using encrypted with a passphrase or is it unencrypted?
3. Do you know what the type and version of the SSH server you are connecting with?
4. Are you able to narrow down which particular release between 0.1.55 and 0.1.62 the problems with the UserAuthPublicKey class started?

Thanks,
Jeremy

norrisjeremy:
Also, if you can run the following nmap command against the server in question, it can help provide details as to what SSH version it is and what crypto algorithms it supports:
```
nmap --script ssh2-enum-algos -sV -p <SSH_PORT> <IP_ADDRESS_OF_SERVER>
```

ihudedi:
Hi @mwiede @avnerw74 
1. I am able to capture debug log via Jsch logger.
2. We are using public key authentication with passphrase
3. Remote version string: SSH-2.0-CerberusFTPServer_9.0
[...]

PR #61, opened 2021-08-12
Several updates

norrisjeremy:
1. Reduce number of algorithms that are runtime checked to improve baseline performance (came up in #47).
2. Add system properties to allow overriding the default crypto algorithms used (in response to #59).
3. Switch from PubkeyAcceptedKeyTypes to PubkeyAcceptedAlgorithms to match recent changes in OpenSSH.
4. Prefer ssh-ed25519 to match recent changes in OpenSSH.
5. Allow configuration of DHGEX sizes and align the default sizes with recommendations from RFC-8270.
6. Directly import JZlib since it appears unmaintained upstream.
7. Add license files for OpenJAX & jBCrypt.
8. Update dependencies to latest versions.
9. Switch Github CI from Java 15 to Java 16 since that is current version of OpenJDK.
10. Modularize jarfile.
11. Mark a few classes package private that are not intended to be public.


norrisjeremy:
FYI, with the Github CI change from Java 15 to 16, the CI checks above probably will hang until you update the project to expect them from Java 16.

Also,  when I was sweeping through dependency updates, I wasn't sure if the `wagon-ssh-external` extension should be updated to a newer version? The version specified in `pom.xml` appears to be quite old.

mwiede:
@norrisjeremy can you please update the Readme of the frontpage with some information on system properties, so that users know, what to to do in order to configure old algorithms/old behavior? Thanks for your contribution!

norrisjeremy:
Yep, I'll update the Readme later today or over the weekend and push an update to this PR.

norrisjeremy:
Let me know if the changes to the Readme look ok, or if you would like additional changes.

PR #68, opened 2021-09-06
Allow ssh-agent keys that are unsupported

norrisjeremy:
* Allow usage of keys from ssh-agent that would otherwise not work because of missing support for the signature algorithm.
* Add integration tests for ssh-agent support.
* Added support for the following algorithms with older Java releases by using [Bouncy Castle](https://www.bouncycastle.org/):
  * ssh-ed25519
  * ssh-ed448
  * curve25519-sha256
  * curve25519-sha256@libssh.org
  * curve448-sha512
  * chacha20-poly1305@openssh.com

FYI, allowing usage of ssh-agent keys allows using ssh-ed25519 from Java versions < 15, since the signature generation process is delegated to the ssh-agent process and does not rely upon Ed25519 support within Java itself.

Also, the ssh-agent integration tests can only be run from Linux, since it relies on sharing a bind mount between the host and container running the ssh-agent process, and it doesn't seem possible to share the ssh-agent's unix domain socket through this bind mount on macOS (and I assume Windows as well).

FYI, the Bouncy Castle addition is completely optional: JSch will still function without it being present in the classpath, albeit without supporting the above mentioned algorithms on older Java releases. This should help address some issues that have been raised by users asking why we don't support these algorithms on older Java releases: all the user has to do is add the `bcprov-jdk15on` jar-file to their classpath to gain support.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -64,6 +64,7 @@ As I explained in a [blog post](http://www.matez.de/index.php/2020/06/22/the-fut
   * This library is a Multi-Release-jar, which means that you can only use certain features when a more recent Java version is used.
     * In order to use ssh-ed25519 & ssh-ed448, you must use at least Java 15.
     * In order to use curve25519-sha256, curve448-sha512 & chacha20-poly1305@<!-- -->openssh.com, you must use at least Java 11.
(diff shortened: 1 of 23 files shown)


mwiede:
Thats awesome! I also had the bouncycastle fallback on my list, but now you have already done it. Thank you

Issue #76, opened 2021-09-13
Host keyword in ssh config should be case-insensitive


jsundmannorberg:
Hi,

According to `man ssh_config` keywords should be case-insensitive and arguments should be case-sensitive. 

However, it appears that the "Host" keyword is case-sensitive in the current implementation (on line 127 in `OpenSSHConfig.java`  we have  a case-sensitive equality check `if(key_value[0].equals("Host")){` )

We have seen cases where this leads to confusing bugs when users assume that the config keywords are case-insensitive, so it would probably be good to change the code to match the specification.

Regards,
Johan Sundman Norberg

norrisjeremy:
Hi @jsundmannorberg,

Are you mostly concerned with treatment of the "Host" keyword itself? Or is handling other keywords important as well?

While fixing the instance for the "Host" keyword you pointed out is a simple one-line change, there are other `ssh_config` type keywords that are are used in other classes (like JSch.java, Session.java, etc.).
This would likely require more signficant changes, such as converting the internal config Map from using `Hashtable<String, String>` to `TreeMap<String, String>` with a `String.CASE_INSENSITIVE_ORDER` type `Comparator<String>`, which would in turn require careful analysis of internal locking, since `TreeMap` isn't thread-safe like `Hashtable`.

Thanks,
Jeremy

jsundmannorberg:
Hi @norrisjeremy,

The Host keyword is the one we have encountered. In this case it is made confusing by the fact that other keywords in the same class are case-insensitive (the `find` method converts all keys to uppercase before searching).

I haven't really looked at how other keywords are treated, but at least they have not caused any issues for us so far.

Regards,
Johan Sundman Norberg

norrisjeremy:
Hi @jsundmannorberg,

The new 0.1.68 release includes a fix to make the "Host" keyword case-insensitive.

Thanks,
Jeremy

norrisjeremy:
Hi @jsundmannorberg,

Well that is an embarrassing blunder. Would something like this look correct to you?

```
diff --git a/src/main/java/com/jcraft/jsch/OpenSSHConfig.java b/src/main/java/com/jcraft/jsch/OpenSSHConfig.java
index a24eb56..8068784 100644
--- a/src/main/java/com/jcraft/jsch/OpenSSHConfig.java
+++ b/src/main/java/com/jcraft/jsch/OpenSSHConfig.java
[...]

PR #79, opened 2021-09-14
More legacy algorithm support and bugfixes

norrisjeremy:
* Added support for the rijndael-cbc@lysator.liu.se algorithm
* Added support for the hmac-ripemd160, hmac-ripemd160@openssh.com and hmac-ripemd160-etm@openssh.com algorithms using [Bouncy Castle](https://www.bouncycastle.org/java.html)
* Added support for various algorithms from [RFC 4253](https://datatracker.ietf.org/doc/html/rfc4253) and [RFC 4344](https://datatracker.ietf.org/doc/html/rfc4344) using [Bouncy Castle](https://www.bouncycastle.org/java.html)
  * cast128-cbc
  * cast128-ctr
  * twofish-cbc
  * twofish128-cbc
  * twofish128-ctr
  * twofish192-cbc
  * twofish192-ctr
  * twofish256-cbc
  * twofish256-ctr
* Added support for the seed-cbc@ssh.com algorithm using [Bouncy Castle](https://www.bouncycastle.org/java.html)
* Address [#76](https://github.com/mwiede/jsch/issues/76) by making the "Host" keyword case-insensitive

### DIFF
--- a/src/main/java/com/jcraft/jsch/bc/CAST128CBC.java
+++ b/src/main/java/com/jcraft/jsch/bc/CAST128CBC.java
@@ -1,0 +1,91 @@
+/* -*-mode:java; c-basic-offset:2; indent-tabs-mode:nil -*- */
+/*
+Copyright (c) 2005-2018 ymnk, JCraft,Inc. All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+  1. Redistributions of source code must retain the above copyright notice,
+     this list of conditions and the following disclaimer.
+
+  2. Redistributions in binary form must reproduce the above copyright 
+     notice, this list of conditions and the following disclaimer in 
+     the documentation and/or other materials provided with the distribution.
+
+  3. The names of the authors may not be used to endorse or promote products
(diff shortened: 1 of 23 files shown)


src/main/java/com/jcraft/jsch/bc/CAST128CBC.java
```diff
@@ -0,0 +1,91 @@
+/* -*-mode:java; c-basic-offset:2; indent-tabs-mode:nil -*- */
+/*
+Copyright (c) 2005-2018 ymnk, JCraft,Inc. All rights reserved.
```
tr00per, line 3: This file is just added. Is it copyrighted by JCraft? Question applies to other freshly added files as well.

From the blog post http://www.matez.de/index.php/2020/06/22/the-future-of-jsch-without-ssh-rsa/ I understood, that JCraft is no longer interested in developing this software. So why new files keep this notice unchanged?

norrisjeremy: It was copied from an existing file authored by JCraft and then lightly modified. Therefore I believe any original copyright notices should remain intact.

Issue #85, opened 2021-10-12
Incompatibility with legacy jsch as used by jgit


ST-DDT:
I tried to replace the original jsch with this updated variant, but that caused an NPE.

````java
CloneCommand cloneCommand = Git.cloneRepository()
                .setURI("git@github.com:mwiede/jsch.git")
                .setDirectory(new File("target/junit/clone-test));
assertDoesNotThrow(cloneCommand::call).close(); // <-- Error
````

This is caused by the following line: [Source](https://gerrit.googlesource.com/jgit/+/refs/tags/v5.13.0.202109080827-r/org.eclipse.jgit.ssh.jsch/src/org/eclipse/jgit/transport/JschConfigSessionFactory.java#401)

````java
JSch.setConfig("ssh-rsa", JSch.getConfig("signature.rsa"));
JSch.setConfig("ssh-dss", JSch.getConfig("signature.dss"));
````

In the original jsch version this was set to:

````java
config.put("signature.rsa", "com.jcraft.jsch.jce.SignatureRSA");
````

This version no longer sets these properties.

My dependencies:

````xml
        <!-- Git -->
        <dependency>
            <groupId>org.eclipse.jgit</groupId>
            <artifactId>org.eclipse.jgit</artifactId>
            <version>5.13.0.202109080827-r</version>
        </dependency>
        <dependency>
            <groupId>org.eclipse.jgit</groupId>
            <artifactId>org.eclipse.jgit.ssh.jsch</artifactId>
            <version>5.13.0.202109080827-r</version>
            <exclusions>
                <exclusion>
                    <groupId>com.jcraft</groupId>
                    <artifactId>jsch</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
        <dependency>
            <groupId>com.github.mwiede</groupId>
            <artifactId>jsch</artifactId>
            <version>0.1.68</version>
        </dependency>
        <dependency>
            <groupId>com.jcraft</groupId>
            <artifactId>jsch.agentproxy.connector-factory</artifactId>
            <version>0.0.9</version>
        </dependency>
        <dependency>
            <groupId>com.jcraft</groupId>
            <artifactId>jsch.agentproxy.jsch</artifactId>
            <version>0.0.9</version>
            <exclusions>
                <exclusion>
                    <groupId>com.jcraft</groupId>
                    <artifactId>jsch</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
````

Stacktrace

````txt
[...]

PR #86, opened 2021-10-13
Fix PageantConnector struct class visibility

ST-DDT:
The visibility of the struct classes need to be public otherwise the following exception will be thrown:

````txt
Caused by: java.lang.IllegalAccessException: Class com.sun.jna.Structure can not access a member of class com.jcraft.jsch.PageantConnector$COPYDATASTRUCT64 with modifiers "public"
	at sun.reflect.Reflection.ensureMemberAccess(Reflection.java:102)
	at java.lang.reflect.AccessibleObject.slowCheckMemberAccess(AccessibleObject.java:296)
	at java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:288)
	at java.lang.reflect.Field.get(Field.java:390)
	at com.sun.jna.Structure.getFieldValue(Structure.java:650)
	... 101 more
````

`protected` seems to work as well, but this might be due to the test setup.

However, this causes these internal data classes to become accessible from the outside.
As a workaround it is also possible to move the data classes inside the private `User32` interface.
Should we do this instead?

See also: https://stackoverflow.com/questions/66214743/instantiating-subclass-of-jna-structure-throws-illegalaccessexception

I have verified the code with gitlab, jgit and KeePass (as Pageant SSHAgent):

https://github.com/mwiede/jsch/issues/85#issuecomment-942086643

### DIFF
--- a/src/main/java/com/jcraft/jsch/PageantConnector.java
+++ b/src/main/java/com/jcraft/jsch/PageantConnector.java
@@ -88,7 +88,7 @@ public class PageantConnector implements AgentConnector {
     long SendMessage(HWND hWnd, int msg, WPARAM num1, byte[] num2);
   }
 
-  private static class COPYDATASTRUCT32 extends Structure {
+  public static class COPYDATASTRUCT32 extends Structure {
     public int dwData;
     public int cbData;
     public Pointer lpData;
@@ -99,7 +99,7 @@ public class PageantConnector implements AgentConnector {
     }
   }
 
-  private static class COPYDATASTRUCT64 extends Structure {
+  public static class COPYDATASTRUCT64 extends Structure {
     public int dwData;
     public long cbData;
     public Pointer lpData;


norrisjeremy:
I also wonder if this means that some changes might be necessary to `module-info.java` in order to allow JNA to perform reflection operations into these classes?

norrisjeremy:
> > I think it should be fine to just make them public. I wonder if we should change the visibility of the User32 interface as well?
> 
> This doesn't seem to be necessary (at least for my tests).

I think it probably wouldn't hurt to go ahead and change it to public visibility, and that would hopefully cover any sort of corner cases that we aren't detecting. Would you mind doing that in this PR?
> 
> > I also wonder if this means that some changes might be necessary to `module-info.java` in order to allow JNA to perform reflection operations into these classes?
> 
> I don't know. Is there an easy way to create a test project that actually uses the modules instead of only defining one?

I think after looking at it further, we should be fine: the `exports com.jcraft.jsch;` should allow reflective access to these classes once their visibility is declared `public`.

PR #90, opened 2021-10-30
Bugfixes

norrisjeremy:
* Address [#89](https://github.com/mwiede/jsch/issues/89) by fixing rare ECDSA signature validation issue
* Address [#93](https://github.com/mwiede/jsch/issues/93) by always setting the "want reply" flag for "env" type channel requests to false
* Update dependencies to latest versions

PR #92, opened 2021-11-04
env passing should be unwant reply

### LINKED ISSUE #93: RequestEnv should be unwant reply
RFC4254 Section 6.4 describe environment variable passing can be filtered in security imprecations.
This can be resulted a reply become false, but it is not an error.

According to JGit discussion about behavior of Jsch env request,
it should be 'unwant-reply' and it is a JSch bug that it throws exception.
https://bugs.eclipse.org/bugs/show_bug.cgi?id=576922

When JGit request GIT_PROTOCOL=version=2 environment variable,
JSch should return even when it is not succeeded.

miurahr:
RFC4254 Section 6.4 describe environment variable passing can be filtered in security imprecations.
This can be resulted a reply become false, but it is not an error.

According to JGit discussion about behavior of Jsch env request,
it should be 'unwant-reply' and its JSch bug.
https://bugs.eclipse.org/bugs/show_bug.cgi?id=576922

When JGit request GIT_PROTOCOL=version=2 environment variable,
JSch should return even when it is not succeeded.

This modification intend to change env request to be unwant-reply.

resolve #93

Signed-off-by: Hiroshi Miura <miurahr@linux.com>

### DIFF
--- a/src/main/java/com/jcraft/jsch/ChannelSession.java
+++ b/src/main/java/com/jcraft/jsch/ChannelSession.java
@@ -219,7 +219,8 @@ class ChannelSession extends Channel{
         byte[] name=_env.nextElement();
         byte[] value=env.get(name);
         request=new RequestEnv();
-        ((RequestEnv)request).setEnv(toByteArray(name), 
+        request.setReply(false);
+        ((RequestEnv)request).setEnv(toByteArray(name),
                                      toByteArray(value));
         request.request(_session, this);
       }
--- a/src/main/java/com/jcraft/jsch/Request.java
+++ b/src/main/java/com/jcraft/jsch/Request.java
@@ -31,17 +31,21 @@ package com.jcraft.jsch;
 
 abstract class Request{
   private boolean reply=false;
+  private boolean checkReply=false;
   private Session session=null;
   private Channel channel=null;
   void request(Session session, Channel channel) throws Exception{
     this.session=session;
     this.channel=channel;
     if(channel.connectTimeout>0){
-      setReply(true);
+      reply=true;
     }
   }
   boolean waitForReply(){ return reply; }
-  void setReply(boolean reply){ this.reply=reply; }
+  void setReply(boolean reply){
+    this.reply=reply;
+    this.checkReply=reply;
+  }
   void write(Packet packet) throws Exception{
     if(reply){
       channel.reply=-1;
@@ -60,8 +64,10 @@ abstract class Request{
           throw new JSchException("channel request: timeout");
         }
       }
-
-      if(channel.reply==0){
+      // When asked unwant-reply/setReply(false) and timeout>0,
+      // we ignore a reply value.
+      // When asked want-reply/setReply(true), we check it.
+      if(checkReply && channel.reply==0){
 	throw new JSchException("failed to send channel request");
       }
     }


norrisjeremy:
Hi @miurahr,

I think a simpler fix for this would be something like this, can you confirm?

```
diff --git a/src/main/java/com/jcraft/jsch/RequestEnv.java b/src/main/java/com/jcraft/jsch/RequestEnv.java
index 02f14f1..ad4afca 100644
--- a/src/main/java/com/jcraft/jsch/RequestEnv.java
+++ b/src/main/java/com/jcraft/jsch/RequestEnv.java
@@ -39,6 +39,7 @@ class RequestEnv extends Request{
   @Override
   public void request(Session session, Channel channel) throws Exception{
     super.request(session, channel);
+    setReply(false);
 
     Buffer buf=new Buffer();
     Packet packet=new Packet(buf);
@@ -47,7 +48,7 @@ class RequestEnv extends Request{
     buf.putByte((byte) Session.SSH_MSG_CHANNEL_REQUEST);
     buf.putInt(channel.getRecipient());
     buf.putString(Util.str2byte("env"));
-    buf.putByte((byte)(waitForReply() ? 1 : 0));
+    buf.putByte((byte) 0);
     buf.putString(name);
     buf.putString(value);
     write(packet);
```

Thanks,
Jeremy

miurahr:
> I think a simpler fix for this would be something like this, can you confirm?

No, it does not works when user set both timeout and environment variables.
When set timeout, `waitForReply()` will return true, and `Request.write()` try checks `reply` value and time.

norrisjeremy:
> > I think a simpler fix for this would be something like this, can you confirm?
> 
> No, it does not works when user set both timeout and environment variables. When set timeout, `waitForReply()` will return true, and `Request.write()` try checks `reply` value and time.

I'm afraid I don't understand what you mean, my proposed fix explicitly calls `setReply(false)`, which sets the `reply` variable to false before it calls `write(packet)`.

norrisjeremy:
Hi @miurahr,

To help illustrate my previous response:

```
diff --git a/src/main/java/com/jcraft/jsch/RequestEnv.java b/src/main/java/com/jcraft/jsch/RequestEnv.java
index 02f14f1..65f4cfa 100644
--- a/src/main/java/com/jcraft/jsch/RequestEnv.java
+++ b/src/main/java/com/jcraft/jsch/RequestEnv.java
@@ -39,6 +39,8 @@ class RequestEnv extends Request{
   @Override
   public void request(Session session, Channel channel) throws Exception{
     super.request(session, channel);
+    //Override and always set `reply=false`
+    setReply(false);
 
     Buffer buf=new Buffer();
     Packet packet=new Packet(buf);
@@ -47,9 +49,11 @@ class RequestEnv extends Request{
     buf.putByte((byte) Session.SSH_MSG_CHANNEL_REQUEST);
     buf.putInt(channel.getRecipient());
     buf.putString(Util.str2byte("env"));
-    buf.putByte((byte)(waitForReply() ? 1 : 0));
+    //Always set want-reply to false (zero)
+    buf.putByte((byte) 0);
     buf.putString(name);
     buf.putString(value);
+    //`reply` has been explicitly set to false before calling `write(packet)`
     write(packet);
   }
 }
```

miurahr:
@norrisjeremy  Does your change checks timeout?

You disable reply by `setReply(false)`, then a chunk

```java
 if(reply){
      long start=System.currentTimeMillis();
      long timeout=channel.connectTimeout;
      while(channel.isConnected() && channel.reply==-1){
	try{Thread.sleep(10);}
	catch(Exception ee){
	}
          if(timeout>0L &&
           (System.currentTimeMillis()-start)>timeout){
          channel.reply=0;
          throw new JSchException("channel request: timeout");
        }
      }
```
is not executed anymore.

norrisjeremy:
> @norrisjeremy Does your change checks timeout?
> 
> You disable reply by `setReply(false)`, then a following part of `Request#write`
> 
> ```java
>  if(reply){
>       long start=System.currentTimeMillis();
>       long timeout=channel.connectTimeout;
>       while(channel.isConnected() && channel.reply==-1){
> 	try{Thread.sleep(10);}
> 	catch(Exception ee){
> 	}
>           if(timeout>0L &&
>            (System.currentTimeMillis()-start)>timeout){
>           channel.reply=0;
>           throw new JSchException("channel request: timeout");
>         }
>       }
> ```
> 
> is not executed anymore.

I don't think there is any need to execute that code sequence: if the `want-reply` flag in the request message is set to zero, that instructs the SSH server to not send a reply back for the request, so there should not be any reason to busy-wait on a reply that is never going to come.

norrisjeremy:
According to RFC 4254 5.4:
```
   If 'want reply' is FALSE, no response will be sent to the request.
   Otherwise, the recipient responds with either SSH_MSG_CHANNEL_SUCCESS
   or SSH_MSG_CHANNEL_FAILURE, or request-specific continuation
   messages.  If the request is not recognized or is not supported for
   the channel, SSH_MSG_CHANNEL_FAILURE is returned.
```

miurahr:
@norrisjeremy Thank you for precise explanation.

I confirms and agreed that your change is enough to fix the problem.

Set environment variable through channel is `unwant-reply` w/o timeout, then git fetch will be requested with timeout check and want-reply.

norrisjeremy:
> @norrisjeremy Thank you for precise explanation.
> 
> I confirms and agreed that your change is enough to fix the problem.
> 
> Set environment variable through channel is `unwant-reply` w/o timeout, then git fetch will be requested with timeout check and want-reply.

Hi @miurahr,

Excellent! Do you want me to submit a PR with my proposed simpler fix? Or would you rather update this PR to use it instead?

Thanks,
Jeremy

miurahr:
> Do you want me to submit a PR with my proposed simpler fix? 

Yes, could you please submit it?

norrisjeremy:
Hi @miurahr,

I have included the proposed fix in #90.

Thanks,
Jeremy

Issue #103, opened 2021-12-14
How to read the Jsch log files


uffeweil:
Hi!
As some of the algorithms has been removed in newer versions of Jsch some of our customers have been reporting troubles in connecting to their servers when upgrading our product (DbVisualizer including upgraded Jsch)

I have seen the system properties (e.g. jsch.kex) that exists to be able to restore old connection capabilities.
Though I have not found any information in how to interpret the logs. 
I.e to be able to instruct our users to set the appropriate properties (or set values in the ssh config file)

The specific case I'm struggling with now is a user running Java 11.0.11 and is getting 
_com.jcraft.jsch.JSchException: Algorithm negotiation fail_
I have attached the log for reference. 

[LogJschSetup.txt](https://github.com/mwiede/jsch/files/7714038/LogJschSetup.txt)

Any help would be highly appreciated

Regards Ulf

mwiede:
Hi @uffeweil 
in your log I see entries `+diffie-hellman-group1-sha1` and `+ssh-dss`, but Jsch does not implement the `+` and `-` like openssh does. That is why the kex guessing fails, because it does not find an algorithm name match including `+`.

Is this coming from an openssh config file? We should be able to implement this in `com.jcraft.jsch.OpenSSHConfig`.

You can try programmatically with `JSch.setConfig("kex", "diffie-hellman-group-exchange-sha1");` or by config of system properties `java ... -Djsch.kex=diffie-hellman-group-exchange-sha1`. Compare with #37 and #40.

Kind regards
Matthias

uffeweil:
Hi Matthias, 
Thanks for the answer.

I will ask our user about this and ask our user to add **-Djsch.kex=diffie-hellman-group-exchange-sha1** to the start of DbVisualizer.

Kind regards
Ulf

uffeweil:
Hi @mwiede, 

Our user just came back. `-Djsch.kex=diffie-hellman-group-exchange-sha1` seem not to have helped. 

Kind Regards 
Ulf

mwiede:
Hi @uffeweil 
sorry, I think I gave wrong advice. 

The meaning of the log file lines with `kex: server` is what the server is offering as list of algorithms for key exchange.
That means, if you set one algorithm via config, you replace the whole list of algorithms, which the **client** offers.
But as I just checked, `diffie-hellman-group-exchange-sha1` is not offered by the server, so there is no matching algorithm between client and server.
[...]

PR #104, opened 2021-12-16
implement openssh config behavior to handle append, prepend and removal of algorithms

mwiede:
as a user was facing the issue in #103 I implemented the openssh config file interpretation, so that Jsch users can use openssh config file without modifiying them. As mentioned in ssh_config man page, the `+` means appending to the comma separated list, `^` means to prepend and `-` means to remove from the list of algorithms.


src/main/java/com/jcraft/jsch/OpenSSHConfig.java
```diff
@@ -223,6 +239,24 @@ private String find(String key) {
...
       */
+
+      if (keysWithListAdoption.contains(key) && value != null && (value.startsWith("+") || value.startsWith("-") || value.startsWith("^"))) {
+
+        String origConfig = JSch.getConfig(originalKey).trim();
+
+        if (value.startsWith("+")) {
+          value=origConfig + "," + value.substring(1);
+        } else if (value.startsWith("-")) {
+          List<String> algList = Arrays.stream(origConfig.split(",")).collect(Collectors.toList());
```
norrisjeremy, line 250: Should this use `Util.split(origConfig, ",")`?

norrisjeremy, line 250: Should this use `Util.split(origConfig, ",")`?

src/main/java/com/jcraft/jsch/OpenSSHConfig.java
```diff
@@ -223,6 +239,24 @@ private String find(String key) {
...
+          for (String alg : value.substring(1).split(",")) {
```
norrisjeremy, line 251: Should this use `Util.split(value.substring(1), ",")`?

src/main/java/com/jcraft/jsch/OpenSSHConfig.java
```diff
@@ -74,6 +77,9 @@
  */
 public class OpenSSHConfig implements ConfigRepository {
 
+  private static final List<String> keysWithListAdoption = Stream.of("KexAlgorithms", "Ciphers", "HostbasedAcceptedAlgorithms",
```
norrisjeremy, line 80: Would it be more appropriate to use a `Set<String>` since the values should be unique?

src/main/java/com/jcraft/jsch/OpenSSHConfig.java
```diff
@@ -74,6 +77,9 @@
  */
 public class OpenSSHConfig implements ConfigRepository {
 
+  private static final List<String> keysWithListAdoption = Stream.of("KexAlgorithms", "Ciphers", "HostbasedAcceptedAlgorithms", 
+  "HostKeyAlgorithms", "MACs", "PubkeyAcceptedAlgorithms", "CASignatureAlgorithms").map(String::toUpperCase).collect(Collectors.toList());
```
norrisjeremy, line 81: Do we need to handle the legacy names like `PubkeyAcceptedKeyTypes` & `HostbasedKeyTypes` that were used until OpenSSH 8.4?

norrisjeremy (commented):
A few notes (see comments).

src/main/java/com/jcraft/jsch/OpenSSHConfig.java
norrisjeremy: And `HostbasedAcceptedKeyTypes` as well?

src/main/java/com/jcraft/jsch/OpenSSHConfig.java
mwiede: yes, I will change it

src/main/java/com/jcraft/jsch/OpenSSHConfig.java
mwiede: I will add PubkeyAcceptedKeyTypes, but HostbasedKeyTypes is not handled anywhere in Jsch

src/main/java/com/jcraft/jsch/OpenSSHConfig.java
```diff
@@ -74,6 +78,11 @@
  */
 public class OpenSSHConfig implements ConfigRepository {
 
+  private static final Set<String> keysWithListAdoption = Stream
+      .of("KexAlgorithms", "Ciphers", "HostbasedAcceptedAlgorithms",
+          "HostKeyAlgorithms", "MACs", "PubkeyAcceptedAlgorithms", "CASignatureAlgorithms", "PubkeyAcceptedKeyTypes")
```
norrisjeremy, line 83: Should we keep the keywords in here that aren't currently supported (like `HostbasedAcceptedAlgorithms` & `CASignatureAlgorithms`)?

mwiede: correct. I removed them

src/main/java/com/jcraft/jsch/OpenSSHConfig.java
```diff
@@ -143,6 +151,15 @@ public Config getConfig(String host) {
     return new MyConfig(host);
   }
 
+  /**
+   * Returns mapping of jsch config property names to OpenSSH property names.
+   *
+   * @return map
+   */
+  public static Hashtable<String, String> getKeymap() {
```
norrisjeremy, line 159: Do we want to mark this as package private so that it doesn't become part of the public facing API?

src/main/java/com/jcraft/jsch/OpenSSHConfig.java
```diff
@@ -223,6 +241,24 @@ private String find(String key) {
         }
       }
       */
+
+      if (keysWithListAdoption.contains(key) && value != null && (value.startsWith("+") || value.startsWith("-") || value.startsWith("^"))) {
+
+        String origConfig = JSch.getConfig(originalKey).trim();
+
+        if (value.startsWith("+")) {
+          value=origConfig + "," + value.substring(1);
```
norrisjeremy, line 250: Should we `trim()` the `value.substring(1)` here?

src/main/java/com/jcraft/jsch/OpenSSHConfig.java
```diff
@@ -223,6 +241,24 @@ private String find(String key) {
...
+        } else if (value.startsWith("-")) {
+          List<String> algList = Arrays.stream(Util.split(origConfig,",")).collect(Collectors.toList());
+          for (String alg : Util.split(value.substring(1),",")) {
```
norrisjeremy, line 253: Should we `trim()` the `value.substring(1)` here?

mwiede: of course

src/main/java/com/jcraft/jsch/OpenSSHConfig.java
mwiede: yes

norrisjeremy (approved):
LGTM.

jthalmai:
It would be nice to have this option also for the configuration with java system properties.

PR #105, opened 2021-12-17
Switch chacha20-poly1305@openssh.com to a pure Bouncy Castle based implementation

norrisjeremy:
Switch chacha20-poly1305@openssh.com to a pure Bouncy Castle based implementation.

JEP 329 implements an RFC-7539 version of ChaCha20 (96 bit nonce + 32
bit counter), while OpenSSH uses a different variant of ChaCha20 (64 bit
nonce + 64 bit counter).

Switching to Bouncy Castle's ChaChaEngine eliminates dependence upon JEP 329's
implementation working for the non RFC-7539 version of ChaCha20 that OpenSSH
utilizes.


norrisjeremy:
This change also eliminates the need to carry a Poly1305 implementation (pulled from OpenJAX) and can instead rely upon using the one distributed by Bouncy Castle.

Issue #75, opened 2021-09-10
Disable ssh-rsa (RSA/SHA1) signature scheme by default


norrisjeremy:
OpenSSH has stated in the release notes for [version 8.7](https://www.openssh.com/txt/release-8.7) that they will finally disable the ssh-rsa (RSA/SHA1) signature scheme by default in their next release (see https://github.com/openssh/openssh-portable/commit/2344750250247111a6c3c6a4fe84ed583a61cc11).

Since SHA-1 is no longer considered cryptographically secure, I think we should follow suit at some point in the near future, so that JSch provides sensible defaults that are aligned with general security practices.

This would not unequivocally disable the use of ssh-rsa type keys themselves, as they can still be utilized with the rsa-sha2-256 (RSA/SHA256) & rsa-sha2-512 (RSA/SHA512) signature schemes.

Additionally users would still be able to programmatically reenable the ssh-rsa (RSA/SHA1) signature scheme if needed via the `server_host_key` & `PubkeyAcceptedAlgorithms` config options (or `jsch.server_host_key` & `jsch.client_pubkey` system properties).

norrisjeremy:
FYI, OpenSSH [version 8.8](https://www.openssh.com/txt/release-8.8) was released last week and includes the change to disable ssh-rsa (RSA/SHA1) signatures by default.

norrisjeremy:
@mwiede What are your thoughts on disabling RSA/SHA1 by default to follow OpenSSH's lead?

mwiede:
@norrisjeremy yes, we should do it with one of the next releases. We should prepare a little more information, that this change might affect many users with old systems.
Maybe we think about making a 0.2.x version to express that "major" change. I always thought about changing the versioning to make it more visible, that this library has its own lifecycle.

norrisjeremy:
Ok, changing the version number to something like 0.2.x sounds like a good idea when we make this change.

mwiede:
Interesting blog post on this topic from github https://github.blog/2021-09-01-improving-git-protocol-security-github/

PR #113, opened 2022-01-21
0.2.0 changes

norrisjeremy:
- Disable RSA/SHA1 signature algorithm by default #75.
- Add basic Logger implementations.
- Fix client version to be compliant with [RFC 4253 section 4.2](https://datatracker.ietf.org/doc/html/rfc4253#section-4.2) by not including minus sign characters #115.
- Add java.util.zip based compression implementation #114.
- Remove deprecated oss-parent #116.
- Use flatten-maven-plugin to produce flattened POM.
- Add JZlib & jBCrypt licenses to POM.
- Include license files in jar file.
- Use maven-assembly-plugin to build sources jar instead of maven-sources-plugin so that Java 9+ sources are includes as well as licenses.
- Only include com.jcraft.jsch package in Javadocs, include Java 9+ sources and reduce verbosity.
- Use Class.asSubclass() method to avoid instantiating an object that isn't descended from the expected interface.
- Follow OpenSSH's lead for skipping packet discard logic when using EtM MACs and also adjust discard lengths to match OpenSSH.


pom.xml
```diff
@@ -318,24 +417,53 @@
...
+            </plugin>
+            <plugin>
+                <groupId>org.sonatype.plugins</groupId>
+                <artifactId>nexus-staging-maven-plugin</artifactId>
+                <version>1.6.8</version>
+                <extensions>true</extensions>
+                <configuration>
+                    <serverId>ossrh</serverId>
+                    <nexusUrl>https://oss.sonatype.org/</nexusUrl>
+                    <autoReleaseAfterClose>true</autoReleaseAfterClose>
```
norrisjeremy, line 463: I'm not sure if you would prefer it to auto release from staging, or to manually release?
If you want to manually release, this would need to changed to `false`.

mwiede:
Hey @norrisjeremy  I am looking through the PR tomorrow

norrisjeremy:
Hi @mwiede,

Great, thanks! Let me know if there are any other changes needed to make your release process work with the removal of `oss-parent` from the POM.

Thanks,
Jeremy

norrisjeremy:
FYI, I have an additional change incoming shortly.

mwiede:
alright, i am waiting with the release

norrisjeremy:
Okie, running one last set of ITs locally, then I'll push a new PR.

Btw, I noticed that you attempted to run release workflow but it didn't work? I'm guessing it's related to the changes I made to remove oss-parent? Do you need assistance to figure out what went wrong?

norrisjeremy:
See #121.

mwiede:
> Okie, running one last set of ITs locally, then I'll push a new PR.
> 
> Btw, I noticed that you attempted to run release workflow but it didn't work? I'm guessing it's related to the changes I made to remove oss-parent? Do you need assistance to figure out what went wrong?

It was the authentication against the sonatype server. I already changed the secret configuration, so it should be fixed with the next one.

norrisjeremy:
Okie, sounds good!

Just curious as to the POM changes in 154d1eeee4b3fd891cf9d1dd51f36e54a8aad471? What sort of issues were you encountering?

mwiede:
no issues per se. I am using this in most of my projects. I find it useful,
because it removes the "conflicts" between transitive dependencies, which
are shown in my "maven helper" plugin in Intellij.

Issue #11, opened 2020-07-17
Critical static analisys issues reported by Klocwork


antoxar:
![image](https://user-images.githubusercontent.com/921451/87797735-66249f80-c84b-11ea-8f09-c73d97c9474b.png)

mwiede:
Yes I am aware of this. I set up Sonarcloud analysis for this repo.
Compare here https://sonarcloud.io/dashboard?id=mwiede_jsch

danielgrahl:
Sonarcloud is currently not analyzing any Java code.

mwiede:
Correct. I deactivated it as it did not work on pr builds.

mwiede:
Sent current code analysis to Sonarcloud. https://sonarcloud.io/project/overview?id=mwiede_jsch

Java Source and Coverage included

PR #124, opened 2022-02-13
Add checkstyle to maven build and fix whitespace after for/while

turbanoff:
JSch is a mature project. As every mature project, it has legacy code with inconsistent formatting. It makes hard to read/modify/audit the source code.
I propose to improve formatting step by step in separate PRs/commits.
In this PR I have added missing spaces after cycles: for/while.
To make sure code style will not regress over time I propose to add `maven-checkstyle-plugin` to build process. Checkstyle has a lot of useful [checks](https://checkstyle.sourceforge.io/checks.html) which can used to verify the code. I have plan to enable some them in separate PRs.
What do you think?

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -1,0 +1,10 @@
+<!DOCTYPE module PUBLIC
+        "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
+        "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
+<module name="Checker">
+    <module name="TreeWalker">
+        <module name="WhitespaceAfter">
+            <property name="tokens" value="LITERAL_WHILE,LITERAL_FOR,DO_WHILE"/>
+        </module>
+    </module>
+</module>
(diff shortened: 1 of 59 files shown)


norrisjeremy:
I have to say that I oppose such a massive change like this.
It would cause severe disruption for several ongoing changes I'm working on, and would also make merging changes from other JSch forks far more difficult.

mwiede:
@turbanoff thanks for your PR. Indeed, clear formatting would help to improve the readability of the code. I will keep it open until  we have finished other things, which might bring more value to the project at the moment. If you look at #11 there are lots of bugs reported from static code analysis, but fixing them might also cause merging problems. Once we decide to decouple from forks and upstream, everything should be open to change those things.

norrisjeremy:
What do we think of instead perhaps moving to the [Google Java Style](https://google.github.io/styleguide/javaguide.html) for everything?

We could just do a single massive reformat of everything once I'm finished with some other outstanding changes.

mwiede:
Closing as we added formatting in #247

Issue #125, opened 2022-02-14
Improved Javadocs


norrisjeremy:
Merge the improved Javadocs from https://github.com/ePaul/jsch-documentation.

norrisjeremy:
FYI, I started working on this today.
It would be helpful if we avoided merge #124 until this is completed.

mwiede:
@norrisjeremy ok for me. Original Jsch code has been there for a while. No rush on the formatting. (or even variable naming :grin:)

PR #126, opened 2022-02-15
Preliminary changes prior to Javadoc work

norrisjeremy:
These are preliminary changes prior to Javadoc improvement targeted for #125.

They are primarily to reduce overall public API surface of the library, by better hiding internal implementation details that users were never intended to access.

Additionally, I've restored some measure of backwards compatibility for several public interfaces that morphed in our fork, by adding default method implementations.

- Restore backwards compatibility of many interfaces by using default method implementations.
- Reduce API surface by making implementation classes non-public.
- Make the IO class non-public since it is only intended for internal use.
- Make the encode method of the Session class non-public since it is only intended for internal use
- Split IdentityRepository.Wrapper class out of IdentityRepository interface since it is only intended for internal use.
- Make abstract algorithm classes non-public.
- Make abstract KeyExchange implementations non-public as they are only intended for internal use.
- Make KeyPair implementations non-public as they can only be meaningfully created via the parent KeyPair class itself.
- Stop implementing Runnable and instead use method references so that run methods are no longer public since they are only intended for internal use


mwiede:
> What are your thoughts on making the `UserAuth` and `KeyExchange` abstract classes non-public? I'm not sure if there are many users that are creating their own `UserAuth` or `KeyExchange` implementations? The advantage if we did make them non-public, is that we could then make the `Buffer` & `Packet` classes non-public, as well as the `Session.read()` & `Session.write()` methods.

in general a good, idea but I see Buffer in `bc` and `jce` package

src/main/java/com/jcraft/jsch/Channel.java
```diff
@@ -32,7 +32,7 @@
 import java.io.*;
 import java.util.Vector;
 
-public abstract class Channel implements Runnable{
+public abstract class Channel{
```
mwiede, line 35: I got your point about internal use. But as I read javadoc of runnable, is it not better to understand, that this instance is supposed to be executed by a thread and keep it as marker interface?

norrisjeremy: My thought behind removing `Runnable` was to be able to change the `run()` method so that it was no longer public, since it isn't intended to be an API method that users would call. If this class implements `Runnable`, then the `run()` method has to be public, as it isn't possible to reduce its visibility.

Also, the primary purpose of `Channel` objects isn't to be a `Runnable` itself, at least from a user's perspective: that's just an internal implementation detail within JSch, with how it handles individual SSH channels within each session.

norrisjeremy:
> in general a good, idea but I see Buffer in `bc` and `jce` package

So I guess we could handle that by moving the `Buffer` class into a separate sub-package, like `com.jcraft.jsch.misc` or `com.jcraft.jsch.util`?

I've also pondered eventually eliminating the `Buffer` class itself someday and instead convert to using `java.nio.ByteBuffer`.

src/main/java/com/jcraft/jsch/Channel.java
mwiede: yes I understand your point. Now its a negotiation whether it is better to hide internals or to show to the developer, that the code of `run()` method is supposed to be run by a `Thread`. I think both ways have their arguments and there is no right or wrong. Let's keep you change for now. Maybe someone else will stumble on this in the future and then we can discuss further.

mwiede:
> java.nio.ByteBuffer

I have not understood why @ymnk implemented his own Buffer. The javadoc says its there since 1.4 so it seems the implementation comes from that old days before 1.4?

Issue #98, opened 2021-12-03
Version.getVersion


thomasrunge:
Any chance to make com.jcraft.jsch.Version.getVersion public? I find it very helpful to ask (and log!) all currently used versions of third party libs.

mwiede:
You could get meta-information from the jar by getting the version also: `com.jcraft.jsch.JSch.class.getPackage().getImplementationVersion()`

norrisjeremy:
> You could get meta-information from the jar by getting the version also: `com.jcraft.jsch.JSch.class.getPackage().getImplementationVersion()`

FYI, I think that only works as long as the original jar-file is used. If you shade the Jsch library into another jar-file (for example using the maven-shade-plugin), I don't think that will work, since the original `MANIFEST.MF` from the JSch jar-file may not be present. This is ultimately why I created this `Version` class a few releases ago, since internally we have an application in which we shade the JSch library and I discovered that the version number wasn't getting set correctly in the SSH connection exchange with a server.

I'm not sure if we should make the `Version` class public, or perhaps just provide a wrapper method in the `JSch` class itself that returns `Version.getVersion()`?

thomasrunge:
Well, the original JSch had JSch.VERSION, you removed that.

norrisjeremy:
Hi @thomasrunge,

Would restoring JSch.VERSION (as proposed in #100) suffice for your use case?

Thanks,
Jeremy

thomasrunge:
That would be perfect, as I'm currently using the old JSch version and switching over to your version would minimize trouble.

mwiede:
@thomasrunge your fix has landed https://search.maven.org/artifact/com.github.mwiede/jsch/0.1.71/jar

Thanks @norrisjeremy for the patch

PR #128, opened 2022-02-17
Issue127

kimmerin:
As "requested" in the issue, these are the changes I've made to be able to set an instance-wide Logger. The static API is kept to avoid breaking issues "out there".

### DIFF
--- a/src/main/java/com/jcraft/jsch/Channel.java
+++ b/src/main/java/com/jcraft/jsch/Channel.java
@@ -45,38 +45,43 @@ public abstract class Channel implements Runnable{
 
   static int index=0; 
   private static Vector<Channel> pool=new Vector<>();
-  static Channel getChannel(String type){
+  static Channel getChannel(String type, Session session){
+    Channel ret = null;
     if(type.equals("session")){
-      return new ChannelSession();
+      ret = new ChannelSession();
     }
     if(type.equals("shell")){
-      return new ChannelShell();
+      ret = new ChannelShell();
     }
     if(type.equals("exec")){
-      return new ChannelExec();
+      ret = new ChannelExec();
     }
     if(type.equals("x11")){
-      return new ChannelX11();
+      ret = new ChannelX11();
     }
     if(type.equals("auth-agent@openssh.com")){
-      return new ChannelAgentForwarding();
+      ret = new ChannelAgentForwarding();
     }
     if(type.equals("direct-tcpip")){
-      return new ChannelDirectTCPIP();
+      ret = new ChannelDirectTCPIP();
     }
     if(type.equals("forwarded-tcpip")){
-      return new ChannelForwardedTCPIP();
+      ret = new ChannelForwardedTCPIP();
     }
     if(type.equals("sftp")){
-      return new ChannelSftp();
+      ret = new ChannelSftp();
     }
     if(type.equals("subsystem")){
-      return new ChannelSubsystem();
+      ret = new ChannelSubsystem();
     }
     if(type.equals("direct-streamlocal@openssh.com")){
-      return new ChannelDirectStreamLocal();
+      ret = new ChannelDirectStreamLocal();
     }
-    return null;
+    if (ret == null) {
+        return null;
+    }
+    ret.setSession(session);
+    return ret;
   }
   static Channel getChannel(int id, Session session){
     synchronized(pool){
@@ -118,7 +123,7 @@ public abstract class Channel implements Runnable{
   volatile int reply=0; 
   volatile int connectTimeout=0;
 
-  private Session session;
+  protected Session session;
 
   int notifyme=0; 
 
--- a/src/main/java/com/jcraft/jsch/Compression.java
+++ b/src/main/java/com/jcraft/jsch/Compression.java
@@ -33,6 +33,9 @@ public interface Compression{
   static public final int INFLATER=0;
   static public final int DEFLATER=1;
   void init(int type, int level) throws Exception;
+  default void init(int type, int level, Session session) throws Exception {
+      init(type, level);
+  }
   byte[] compress(byte[] buf, int start, int[] len);
   byte[] uncompress(byte[] buf, int start, int[] len);
 }
--- a/src/main/java/com/jcraft/jsch/JSch.java
+++ b/src/main/java/com/jcraft/jsch/JSch.java
@@ -280,13 +280,14 @@ public class JSch{
 
   private HostKeyRepository known_hosts=null;
 
-  private static final Logger DEVNULL=new Logger(){
+  static final Logger DEVNULL=new Logger(){
       @Override
       public boolean isEnabled(int level){return false;}
       @Override
       public void log(int level, String message){}
     };
   static Logger logger=DEVNULL;
+  private Logger instLogger;
 
   public JSch(){
   }
@@ -673,7 +674,8 @@ public class JSch{
   /**
    * Sets the logger
    *
-   * @param logger logger
+   * @param logger logger or <code>null</code> if no logging
+   * should take place
    *
    * @see com.jcraft.jsch.Logger
    */
@@ -681,7 +683,33 @@ public class JSch{
     if(logger==null) logger=DEVNULL;
     JSch.logger=logger;
   }
-
+  
+  /**
+   * Returns  a logger to be used for this particular instance of JSch
+   * @return The logger that is used by this instance. If no particular
+   * logger has been set, the statically set logger is returned.
+   */
+  public Logger getInstanceLogger() {
+      if (this.instLogger == null) {
+          return logger;
+      }
+      return instLogger;
+  }
+  
+  /**
+   * Sets a logger to be used for this particular instance of JSch
+   * @param logger The logger to be used or <code>null</code> if
+   * the statically set logger should be used
+   */
+  public void setInstanceLogger(Logger logger) {
+      this.instLogger = logger;
+  }
+  
+  /**
+   * Returns the statically set logger, i.e. the logger being
+   * used by all JSch instances without explicitly set logger.
+   * @return The logger
+   */
   public static Logger getLogger(){
     return logger;
   }
--- a/src/main/java/com/jcraft/jsch/KeyExchange.java
+++ b/src/main/java/com/jcraft/jsch/KeyExchange.java
@@ -69,6 +69,11 @@ public abstract class KeyExchange{
 
   public abstract void init(Session session, 
                             byte[] V_S, byte[] V_C, byte[] I_S, byte[] I_C) throws Exception;
+  void doInit(Session session, 
+                            byte[] V_S, byte[] V_C, byte[] I_S, byte[] I_C) throws Exception {
+    this.session = session;
+    init(session, V_S, V_C, I_S, I_C);
+  }
   public abstract boolean next(Buffer buf) throws Exception;
 
   public abstract int getState();
@@ -96,13 +101,13 @@ public abstract class KeyExchange{
     Buffer sb=new Buffer(I_S); sb.setOffSet(17);
     Buffer cb=new Buffer(I_C); cb.setOffSet(17);
 
-    if(JSch.getLogger().isEnabled(Logger.INFO)){
+    if(session.getLogger().isEnabled(Logger.INFO)){
       for(int i=0; i<PROPOSAL_MAX; i++){
-        JSch.getLogger().log(Logger.INFO,
+        session.getLogger().log(Logger.INFO,
                              "kex: server: "+Util.byte2str(sb.getString()));
       }
       for(int i=0; i<PROPOSAL_MAX; i++){
-        JSch.getLogger().log(Logger.INFO,
+        session.getLogger().log(Logger.INFO,
                              "kex: client: "+Util.byte2str(cb.getString()));
       }
       sb.setOffSet(17);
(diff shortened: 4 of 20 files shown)


src/main/java/com/jcraft/jsch/Compression.java
```diff
@@ -32,7 +32,7 @@
 public interface Compression{
   static public final int INFLATER=0;
   static public final int DEFLATER=1;
-  void init(int type, int level) throws Exception;
+  void init(Session session, int type, int level) throws Exception;
```
norrisjeremy, line 35: In #126 I've attempted to restore the backwards incompatibilities I introduced in several public interfaces.

Could we change revert this method signature back, and then instead add something like this?
```
default void init(int type, int level, Session session) throws Exception {init(type, level);}
default Logger getLogger() {return JSch.getLogger();}
```

That way this interface maintains backwards compatibility.
We can thin change the built-in implementations to override these methods as appropriate, but any users that are using custom `Compression` implementations don't suddenly have their implementations break?

src/main/java/com/jcraft/jsch/JSch.java
```diff
@@ -38,7 +38,7 @@
   /**
    * The version number.
    */
-  public static final String VERSION  = Version.getVersion();
+  public static final String VERSION  = "0.2.0"; //Version.getVersion();
```
norrisjeremy, line 41: This part needs to be reverted.

src/main/java/com/jcraft/jsch/KeyPair.java
```diff
@@ -175,6 +175,15 @@ public void writePrivateKey(OutputStream out, byte[] passphrase){
 
   abstract byte[] getKeyTypeName();
   public abstract int getKeyType();
+  
+  public String getKeyTypeName(String encoding) {
```
norrisjeremy, line 179: This part seems unrelated to the logging changes?

src/main/java/com/jcraft/jsch/PortWatcher.java
```diff
@@ -211,7 +212,9 @@ public void run(){
           channel.connect(connectTimeout);
         } else {
           ChannelDirectTCPIP channel = new ChannelDirectTCPIP();
+          channel.setSession(session);
           channel.init();
+          channel.setSession(session);
```
norrisjeremy, line 217: Why do we have to call `setSession()` twice here (both before and after `init()`?

src/main/java/com/jcraft/jsch/Session.java
```diff
@@ -73,7 +73,7 @@
   private static final int PACKET_MAX_SIZE = 256 * 1024;
 
   private byte[] V_S;                                 // server version
-  private byte[] V_C=Util.str2byte("SSH-2.0-JSCH_"+Version.getVersion()); // client version
+  private byte[] V_C=Util.str2byte("SSH-2.0-JSCH_"+JSch.VERSION); // client version
```
norrisjeremy, line 76: This part needs to be reverted I think.

src/main/java/com/jcraft/jsch/Session.java
```diff
@@ -514,14 +514,17 @@ public void connect(int connectTimeout) throws JSchException{
...
         }
-        if(auth_cancel)
-          throw new JSchException("Auth cancel");
-        throw new JSchException("Auth fail");
+        throw new JSchException((auth_cancel ? "Auth cancel"
+                : "Auth fail")
+                + " for methods '" + smethods + "'");
+//        if(auth_cancel)
+//          throw new JSchException("Auth cancel");
+//        throw new JSchException("Auth fail");
```
norrisjeremy, line 527: I think we should just remove these commented out lines.

src/main/java/com/jcraft/jsch/KeyExchange.java
```diff
@@ -69,6 +69,11 @@
 
   public abstract void init(Session session, 
                             byte[] V_S, byte[] V_C, byte[] I_S, byte[] I_C) throws Exception;
+  public void doInit(Session session,
```
norrisjeremy, line 72: I don't think we should make this public. I'd change it to package private.

norrisjeremy (changes requested):
Here are my initial thoughts/comments.

src/main/java/com/jcraft/jsch/JSch.java
kimmerin: That part wasn't intended to be in the commit

src/main/java/com/jcraft/jsch/KeyPair.java
kimmerin: It is. Shouldn't be part of the commit, either.

src/main/java/com/jcraft/jsch/PortWatcher.java
kimmerin: Shouldn't be. I'll remove the second one (I suppose while going over the changes I've missed the first setSession I've already added and added the second one).

src/main/java/com/jcraft/jsch/Session.java
kimmerin: It's unrelated to the logging issue but worth thinking about the reason why the "logic" to get the implementation's version needs to be done twice.

src/main/java/com/jcraft/jsch/Session.java
kimmerin: This one is logging unrelated and is one of the changes I've done to the JSch sources. Seems that I've copied that to this file together with the logging related changes. If it's OK to keep the "for methods"-part in the exception I can remove the commented out code, otherwise I can revert it to the less informative original message

src/main/java/com/jcraft/jsch/KeyExchange.java
kimmerin: I've changed it. I've copied init's method signature to make sure that the parameters are identical.

src/main/java/com/jcraft/jsch/Session.java
norrisjeremy: This probably could remain. Calling `Version.getVersion()` here originated because earlier I mistakenly removed `JSch.VERSION`, but then re-added it to restore backwards compatibility in #98.

src/main/java/com/jcraft/jsch/Session.java
norrisjeremy: Yes, I think keeping the "for methods" part in the exception is fine. We should just remove the old commented out code here.

src/main/java/com/jcraft/jsch/Compression.java
kimmerin: I've reverted the change but solved it a bit differently to keep code duplication in juz.Compression and jzlib.Compression to a minimum. There is now a default setSession-method doing nothing and a static logMessage in Session with a Supplier<String> that is now called in both Compression-implementations.

src/main/java/com/jcraft/jsch/Session.java
kimmerin: Done

src/main/java/com/jcraft/jsch/Compression.java
```diff
@@ -32,7 +32,10 @@
 public interface Compression{
   static public final int INFLATER=0;
   static public final int DEFLATER=1;
-  void init(Session session, int type, int level) throws Exception;
+  default void setSession(Session session) {
```
norrisjeremy, line 35: I'd rather go with the approach of adding a `default void init(int type, int level, Session session) {init(type, level);}` approach, and let actual implementations of `Compression` handle overriding it.

We would then change the `Session` class to call the new `init(type, level, session)` method.
Any existing implementations that users have developed that don't specialize this default implementation continue to work.
And since the implementations that specialize the new `void init(int type, int level, Session session)` method would then internally gain access to the `Session` instance and could just call the `session.getLogger()` method for their internal logging needs.
That would also allow us to avoid having to add the unwieldy static `logMessage()` method to the `Session` class.

norrisjeremy (changes requested):

src/main/java/com/jcraft/jsch/Compression.java
kimmerin: Is it no valid use case that the Compression classes are used independently, i.e. in an application that need ZLIB compression for something else? In that case your suggestion would break this application because there is no session instance and the logging attempt would lead to a NullPointerException. That's why I've created a logMessage-method that checks for the existence of the session and falls back to the static logger if it's not there.

norrisjeremy: Users can create their own `Compression` implementations and then instruct JSch to use utilize them by calling `JSch.setConfig("zlib@openssh.com", "my.own.CompressionImpl")` or `session.setConfig("zlib@openssh.com", "my.ownCompressionImpl")`.

For the built-in implementations we provide (`com.jcraft.jsch.jzlib.Compression` & `com.jcraft.jsch.juz.Compression`), there wouldn't be an NPE thrown because `Session` would call the new `init(type, level, this);` (thus never calling `init(type, level)`).

norrisjeremy: The built-in implementations of `com.jcraft.jsch.jzlib.Compression` & `com.jcraft.jsch.juz.Compression` aren't intended to be consumed outside of JSch, so I'm not especially if somebody got an NPE in that case, because I don't consider them part of the public API.
In fact, in the `module-info.java` I constructed, these packages (`com.jcraft.jsch.jzlib` & `com.jcraft.jsch.juz`) aren't even exported, to further emphasize that point.

I'm far more interested in allow users to continue to instruct JSch to use their own custom `Compression` implementations without having to adapt them after this change.

norrisjeremy: We can also introduce `null` guards in the `com.jcraft.jsch.jzlib.Compression` & `com.jcraft.jsch.juz.Compression` for when `session` is dereferenced to avoid throwing an NPE, for anybody that does happening to utilize them whilst only calling `init(type, level)`.

src/main/java/com/jcraft/jsch/UserInfo.java
```diff
@@ -36,4 +36,5 @@
   boolean promptPassphrase(String message);
   boolean promptYesNo(String message);
   void showMessage(String message);
+
```
norrisjeremy, line 39: I think this stray newline should be removed?

src/main/java/com/jcraft/jsch/Channel.java
```diff
@@ -45,38 +45,43 @@
 
   static int index=0; 
   private static Vector<Channel> pool=new Vector<>();
-  static Channel getChannel(String type){
+  static Channel getChannel(String type, Session session){
+    Channel ret = null;
     if(type.equals("session")){
-      return new ChannelSession();
+      ret = new ChannelSession();
```
norrisjeremy, line 51: I wonder if we should perhaps we just add a `Session session` argument to each of the `ChannelXXX` implementations instead of adding an explicit `setSession(session)` call below?
That would then allow us to remove the `addChannel()` method in `Session` & `PortWatcher` classes I think?
@mwiede any thoughts?

norrisjeremy (changes requested):

norrisjeremy:
I am wondering: should we structure this such that users have the ability to set a different logger per `Session` created?
I could imagine a case in which a user opens multiple `Session`'s, all to different hosts, all from the same underlying `JSch` instance?
I.e.
```
JSch jsch = new JSch();
Session sess1 = jsch.getSession(username, host1, port);
Session sess2 = jsch.getSession(username, host2, port);
...
```
Some users may desire the ability to have different loggers for `sess1` & `sess2` above?

kimmerin:
> I am wondering: should we structure this such that users have the ability to set a different logger per `Session` created?

I've changed the Session to have its own logger instance that get instantiated by JSch's instance logger and allow to change it by providing a corresponding setter method. That means that you can't change the logger on the JSch-level and have that affect existing sessions but I'm not sure if that should be considered an issue.

BTW: Session contains an "active" System.err.println ("fail in inflater"). Should that be commented out or changed to a logging message?

src/main/java/com/jcraft/jsch/Compression.java
kimmerin: I've done the changes you've suggested including checking for null and fall back to the jsch logger if no particular session logger is specified.

src/main/java/com/jcraft/jsch/jzlib/Compression.java
```diff
@@ -37,31 +39,38 @@
   private Inflater inflater;
   private byte[] tmpbuf=new byte[BUF_SIZE];
   private byte[] inflated_buf;
+  private Session session;
 
-  public Compression(){
+  public Compression() {
```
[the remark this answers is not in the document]
kimmerin: Done

norrisjeremy:
Hi @kimmerin,

I'll try to review again later this evening or over the weekend.

Thanks!
Jeremy

kimmerin:
> I'll try to review again later this evening or over the weekend.

No problem, take your time.

src/main/java/com/jcraft/jsch/juz/Compression.java
```diff
@@ -29,10 +29,22 @@
...
   private byte[] tmpbuf=new byte[BUF_SIZE];
   private byte[] inflated_buf;
+  private Session session;
 
   public Compression(){
   }
 
+  private void logMessage(int level, String message) {
+    Logger logger = session == null ? JSch.getLogger() : session.getLogger();
+    logger.log(level, message);
```
norrisjeremy, line 39: I think we should add a check to make sure `logger.isEnabled(level)` here before actually logging the message?

norrisjeremy (changes requested):

src/main/java/com/jcraft/jsch/juz/Compression.java
norrisjeremy: This would also apply to the `jzlib` implementation of `Compression` as well.

kimmerin: That wouldn't make much sense, because the common reason to do this is to avoid concatenating a log-message that isn't used later on if `level > currentLevel`. In `logMessage` that concatenation has already taken place and `logger.logMessage` is supposed to do this check itself, so doing this check before doesn't save anything.

norrisjeremy: If a user has a custom Logger that has a specific logging level disabled, we shouldn't emit a log message at said level.

kimmerin: I've added a check and used a Supplier, so there is no need for all these `if(enabled)-logMessage`-cascades.

norrisjeremy:
I think this looks good now: can we squash all the commits so that it's one single atomic changeset?

kimmerin:
> can we squash all the commits so that it's one single atomic changeset?

I've tried but have to admit that my knowledge of git isn't very much and following a description I've found online ended with my client trying to push to master and not the branch which is not what should happen, I suppose.

Is there a description somewhere how to do that (ideally with Eclipse and not the git command shell)?

kimmerin:
Hm, it looks like that commit [55e3ef3](https://github.com/mwiede/jsch/pull/128/commits/55e3ef390e5e4ca87c366899a728531dc5ffe1a9)  is the squashed commit that you wanted but then Eclipse did some merge afterwards. Can you use [55e3ef3](https://github.com/mwiede/jsch/pull/128/commits/55e3ef390e5e4ca87c366899a728531dc5ffe1a9) for your merge?

kimmerin:
@norrisjeremy Shall I create a new branch, put all the changes in there, commit them and create a new pull request? That would have the same effect.

mwiede:
@kimmerin you do not need to create a new branch, I can select "squash and merge" which aggregates all commits to one.

kimmerin:
@mwiede I shouldn't write tests ;-) Is the Slf4jLogger set by the test framework during build or where is it coming from?

norrisjeremy:
> @mwiede I shouldn't write tests ;-) Is the Slf4jLogger set by the test framework during build or where is it coming from?

I wonder if you need to add a `@BeforeEach` in `SessionTest` to explicitly reset the JSch static logger instance like in `JSchTest`?

kimmerin:
> I wonder if you need to add a `@BeforeEach` in `SessionTest` to explicitly reset the JSch static logger instance like in `JSchTest`?

Except in my test method there is no call of `setLogger` in this test, so the Slf4JLogger must have been set from the outside. I'm now setting the static logger to `null` at the beginning of the test.

Issue #129, opened 2022-02-21
Allow the configuration of the client version string


kimmerin:
Hi,

many security audits done on server software requires you to deactivate version informations in protocols like HTTP or SSH. To be able to fulfill this requirement it should be possible to configure a client version string to be used. If no particular setting has taken place, the default `"JSCH_" + JSch.Version` should be used.

If accepted I can do the change and do a pull request.

norrisjeremy:
Can't this already be accomplished with the `Session.setClientVersion()` method?

kimmerin:
> Can't this already be accomplished with the `Session.setClientVersion()` method?

To be honest, I've missed this one ;-) Problem with this method is that you have to set everything, including the prefix `SSH-2.0-` and I was more thinking of being able to set it "globally" in the JSch-instance.

kimmerin:
I can add a check if the given String starts with `SSH-2.0-` and if not autoappend it.

norrisjeremy:
My only concern would be if a user somehow relied on the behavior to fully populate the client version string with something non-RFC compliant, in order to appease some custom (non-compliant) SSH server implementation they were interoperating with.

kimmerin:
That's true, didn't think about that.

norrisjeremy:
My other worry about trying to auto-detect whether to prepend the `SSH-2.0-` to their supplied string is it would lead to some inconsistency in the `getClientVersion()` string, in that what a user potentially supplied to `setClientVersion()` may not match what they get back from `getClientVersion()` (since we would be adding the `SSH-2.0-` prefix onto it).

kimmerin:
I've already accepted my autoprefix-idea not being the best ;-)

PR #130, opened 2022-03-09
setClientVersion-implementation as discussed in #129

kimmerin:
Compared to the logger issue this is a more simple change. I've added a system-property that allows to set the client version that way as well and added tests to check the behavior. This required the setup of the config-hashtable to be in a dedicated method.

### DIFF
--- a/src/main/java/com/jcraft/jsch/JSch.java
+++ b/src/main/java/com/jcraft/jsch/JSch.java
@@ -32,6 +32,7 @@ package com.jcraft.jsch;
 import java.io.InputStream;
 import java.util.Enumeration;
 import java.util.Hashtable;
+import java.util.Map;
 import java.util.Vector;
 
 public class JSch{
@@ -42,7 +43,12 @@ public class JSch{
 
   static Hashtable<String, String> config=new Hashtable<>();
   static{
+    fillConfig(config, JavaVersion.getVersion());
+  }
+
+  static void fillConfig(Map<String, String> config, int javaVersion) {
     config.put("kex", Util.getSystemProperty("jsch.kex", "curve25519-sha256,curve25519-sha256@libssh.org,ecdh-sha2-nistp256,ecdh-sha2-nistp384,ecdh-sha2-nistp521,diffie-hellman-group-exchange-sha256,diffie-hellman-group16-sha512,diffie-hellman-group18-sha512,diffie-hellman-group14-sha256"));
+    config.put("client_version", Util.getSystemProperty("jsch.client_version", "SSH-2.0-JSCH_" + VERSION));
     config.put("server_host_key", Util.getSystemProperty("jsch.server_host_key", "ssh-ed25519,ecdsa-sha2-nistp256,ecdsa-sha2-nistp384,ecdsa-sha2-nistp521,rsa-sha2-512,rsa-sha2-256"));
     config.put("prefer_known_host_key_types", Util.getSystemProperty("jsch.prefer_known_host_key_types", "yes"));
     config.put("enable_server_sig_algs", Util.getSystemProperty("jsch.enable_server_sig_algs", "yes"));
@@ -286,6 +292,7 @@ public class JSch{
       @Override
       public void log(int level, String message){}
     };
+  String clientVersion;
   static Logger logger=DEVNULL;
   private Logger instLogger;
 
@@ -712,5 +719,26 @@ public class JSch{
    */
   public static Logger getLogger(){
     return logger;
+  }
+
+  /**
+   * Returns the client version to be used when connecting to a peer
+   * @return the client version
+   */
+  public String getClientVersion() {
+    return clientVersion == null ? getConfig("client_version") : clientVersion;
+  }
+
+  /**
+   * Sets the client version to be used when connecting to a peer. If
+   * <code>null</code> is provided, the version string being set with the
+   * system property <code>jsch.client_version</code> is used. If that
+   * isn't set the default text <code>SSH-2.0-JSCH_[version]</code>
+   * is used. The set text is used as is, i.e. no <code>SSH-2.0-</code>
+   * is prefixed if that part is missing.
+   * @param clientVersion the client version to set
+   */
+  public void setClientVersion(String clientVersion) {
+    this.clientVersion = clientVersion;
   }
 }
--- a/src/main/java/com/jcraft/jsch/JSch.java
+++ b/src/main/java/com/jcraft/jsch/JSch.java
@@ -211,14 +211,14 @@ public class JSch{
 
     config.put("pbkdf", "com.jcraft.jsch.jce.PBKDF");
 
-    if(JavaVersion.getVersion()>=11){
+    if(javaVersion>=11){
       config.put("xdh", "com.jcraft.jsch.jce.XDH");
     }
     else{
       config.put("xdh", "com.jcraft.jsch.bc.XDH");
     }
 
-    if(JavaVersion.getVersion()>=15){
+    if(javaVersion>=15){
       config.put("keypairgen.eddsa", "com.jcraft.jsch.jce.KeyPairGenEdDSA");
       config.put("ssh-ed25519", "com.jcraft.jsch.jce.SignatureEd25519");
       config.put("ssh-ed448", "com.jcraft.jsch.jce.SignatureEd448");
--- a/src/main/java/com/jcraft/jsch/Session.java
+++ b/src/main/java/com/jcraft/jsch/Session.java
@@ -81,7 +81,7 @@ public class Session{
   private static final int PACKET_MAX_SIZE = 256 * 1024;
 
   private byte[] V_S;                                 // server version
-  private byte[] V_C; //=Util.str2byte("SSH-2.0-JSCH_"+Version.getVersion()); // client version
+  private String V_C; //=Util.str2byte("SSH-2.0-JSCH_"+Version.getVersion()); // client version
 
   private byte[] I_C; // the payload of the client's SSH_MSG_KEXINIT
   private byte[] I_S; // the payload of the server's SSH_MSG_KEXINIT
@@ -259,10 +259,11 @@ public class Session{
 
       jsch.addSession(this);
 
+      byte[] clientVersion = Util.str2byte(getClientVersion());
       {
         // Some Cisco devices will miss to read '\n' if it is sent separately.
-        byte[] foo=new byte[V_C.length+2];
-        System.arraycopy(V_C, 0, foo, 0, V_C.length);
+        byte[] foo=new byte[clientVersion.length+2];
+        System.arraycopy(clientVersion, 0, foo, 0, clientVersion.length);
         foo[foo.length-2]=(byte)'\r';
         foo[foo.length-1]=(byte)'\n';
         io.put(foo, 0, foo.length);
@@ -315,7 +316,7 @@ public class Session{
         getLogger().log(Logger.INFO,
                              "Remote version string: "+_v_s);
         getLogger().log(Logger.INFO,
-                             "Local version string: "+Util.byte2str(V_C));
+                             "Local version string: "+Util.byte2str(clientVersion));
       }
 
       send_kexinit();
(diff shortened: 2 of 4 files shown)


src/main/java/com/jcraft/jsch/Session.java
```diff
@@ -2716,9 +2716,13 @@ public String getServerVersion(){
     return Util.byte2str(V_S);
   }
   public String getClientVersion(){
-    return Util.byte2str(V_C);
+    return V_C == null ? jsch.getClientVersion() : Util.byte2str(V_C);
   }
   public void setClientVersion(String cv){
+    V_C = null;
```
norrisjeremy, line 2722: This will lead to an NPE if a user explicitly calls `setClientVersion(null)` because the `V_C` property is dereferenced in the `connect()` & `receive_kexinit()` methods.

norrisjeremy (changes requested):

src/main/java/com/jcraft/jsch/Session.java
kimmerin: All the tests that actually show that require a Docker-container I don't have. I've changed it so that everything is using the get-method now

src/main/java/com/jcraft/jsch/Session.java
```diff
@@ -624,7 +625,7 @@ private KeyExchange receive_kexinit(Buffer buf) throws Exception {
       throw new JSchException(e.toString(), e);
     }
 
-    kex.doInit(this, V_S, V_C, I_S, I_C);
+    kex.doInit(this, V_S, Util.str2byte(getClientVersion()), I_S, I_C);
```
norrisjeremy, line 628: If a user changed the client version string after a Session is connected and a rekeying operation occurred, their SSH connection would then break, because the client version is supposed to be idempotent.

Whilst this could already happen prior, with how you've adapted the Session class to fallback to calling an instance method in the JSch class, we would be making this scenario easier for a user to trigger, since they may naively believe they can change the client version of the JSch instance between multiple calls to `getSession()`.

I'm thinking we should avoid this scenario, by simply initializing the `V_C` property to whatever the value of `jsch.getClientVersion()` is at the time the Session instance is created and avoid trying the callbacks to `jsch.getClientVersion()` within `session.getClientVersion()`.

We simply would need to document to users that the client version of a Session instance is not affected by changes to the JSch instance after the Session is created, and that the only way to manipulate afterwards is by direct calls to `session.setClientVersion()`.

norrisjeremy (changes requested):

src/main/java/com/jcraft/jsch/JSch.java
```diff
@@ -42,7 +43,12 @@
 
   static Hashtable<String, String> config=new Hashtable<>();
   static{
+    fillConfig(config, JavaVersion.getVersion());
+  }
+
+  static void fillConfig(Map<String, String> config, int javaVersion) {
```
norrisjeremy, line 49: What's the purpose of the `javaVersion` argument? It appears to be unused?

norrisjeremy (changes requested):

src/main/java/com/jcraft/jsch/JSch.java
kimmerin: It allows to test the different behavior in dependence of different java versions without the requirement to actually use that particular java version when executing the test

kimmerin: What's missing is a checkFillConfiig in JschTest but that's a big TODO I haven't had the time to attack

src/main/java/com/jcraft/jsch/Session.java
kimmerin: Changing the client version while being in an active session was possible before so we've found a possible bug here as far as I can see it. I've solved it now by recreating a member V_C that is filled with the bytes of the client version that is returned by `getClientVersion` at the time of connect.

src/main/java/com/jcraft/jsch/Session.java
```diff
@@ -81,7 +81,8 @@
   private static final int PACKET_MAX_SIZE = 256 * 1024;
 
   private byte[] V_S;                                 // server version
-  private byte[] V_C=Util.str2byte("SSH-2.0-JSCH_"+JSch.VERSION); // client version
+  private byte[] V_C; // will be set during connect so that the client version
```
norrisjeremy, line 84: Why not simply do `VC=Util.str2byte(jsch.getClientVersion())` here?
Then we could omit all the other changes to this class.

src/main/java/com/jcraft/jsch/JSch.java
kimmerin: ... and the actual use of the parameter was missing as well. Not sure what happened here, I've definitely used it when doing the implementation.

src/main/java/com/jcraft/jsch/Session.java
kimmerin: There might be a setClientVersion on Jsch-level after creating a session and before doing a connect that would get lost that way

norrisjeremy: I think it's easier to document and understand that calls to `jsch.setClientVersion()` have no impact to any previously created Session instances than it is to document and understand that `jsch.setClientVersion()` affects any previously created Session instance that has not yet execute `session.connect()`.

Basically: after you create a Session object, via `jsch.getSession()`, the only way to manipulate the client version string of said Session object should be via `session.setClientVersion()`, independent of any calls to `jsch.setClientVersion()` that occur after you have created the Session object.

src/main/java/com/jcraft/jsch/JSch.java
norrisjeremy: I'm still not sure I understand why we need a `javaVersion` argument instead of simply using `JavaVersion.getVersion()`?
The value returned by `JavaVersion.getVersion()` doesn't dynamically change at runtime.

kimmerin: I've committed JschTest showing the use of this parameter as part of a unit test. The fact that JavaVersion.getVersion doesn't change at runtime is the actual problem if you want to test behavior that depends on this value

src/test/java/com/jcraft/jsch/JSchTest.java
```diff
@@ -64,6 +97,44 @@ void checkLoggerBehavior() throws Exception {
...
+      try {
+        System.setProperties(props);
+        
+        HashMap<String, String> map = new HashMap<>();
+        
+        JSch.fillConfig(map, 10);
+        assertEquals("com.jcraft.jsch.bc.XDH", map.get("xdh"), "check of xdh failed");
+        JSch.fillConfig(map, 11);
+        assertEquals("com.jcraft.jsch.jce.XDH", map.get("xdh"), "check of xdh failed");
+        JSch.fillConfig(map, 12);
```
norrisjeremy, line 116: Do we need to test this for `12` since it is the same as `11`?

src/test/java/com/jcraft/jsch/JSchTest.java
```diff
@@ -64,6 +97,44 @@ void checkLoggerBehavior() throws Exception {
...
+        
+        JSch.fillConfig(map, 14);
+        assertEquals("com.jcraft.jsch.bc.KeyPairGenEdDSA", map.get("keypairgen.eddsa"), "check of keypairgen.eddsa failed");
+        assertEquals("com.jcraft.jsch.bc.SignatureEd25519", map.get("ssh-ed25519"), "check of ssh-ed25519 failed");
+        assertEquals("com.jcraft.jsch.bc.SignatureEd448", map.get("ssh-ed448"), "check of ssh-ed448 failed");
+        JSch.fillConfig(map, 15);
+        assertEquals("com.jcraft.jsch.jce.KeyPairGenEdDSA", map.get("keypairgen.eddsa"), "check of keypairgen.eddsa failed");
+        assertEquals("com.jcraft.jsch.jce.SignatureEd25519", map.get("ssh-ed25519"), "check of ssh-ed25519 failed");
+        assertEquals("com.jcraft.jsch.jce.SignatureEd448", map.get("ssh-ed448"), "check of ssh-ed448 failed");
+        JSch.fillConfig(map, 16);
```
norrisjeremy, line 127: Do we need to test this for `16` since it is the same as `15`?

src/test/java/com/jcraft/jsch/JSchTest.java
kimmerin: I'm always testing the edge cases to make sure that if somebody for some reason changes `>=` to `==`, the test fails.

src/test/java/com/jcraft/jsch/JSchTest.java
kimmerin: Same reason as with the test for xdh

src/test/java/com/jcraft/jsch/JSchTest.java
```diff
@@ -1,17 +1,29 @@
...
 
 class JSchTest {
+    private Hashtable<String, String> orgConfig;
+    private Logger orgLogger;
 
     @BeforeEach
+    @SuppressWarnings("unchecked")
     void resetJsch() {
-        JSch.setLogger(null);
+      orgConfig = (Hashtable<String, String>) JSch.config.clone();
```
norrisjeremy, line 18: How about simply doing `new Hashtable<>(JSch.config)`, to avoid the ugly cast and call to `clone()`?

kimmerin: The "copy-constructor" creates shallow copies but with immutable values this isn't an issue here, so I've changed it as suggested.

src/main/java/com/jcraft/jsch/Session.java
kimmerin: Even if we implement it the way you suggested you need to either have a distinct member for the client version bytes used during a connection or check the state of the session and throw an exception if you try to set the version string after a connection has been established.

Otherwise you will run into the problem you've described before:

> If a user changed the client version string after a Session is connected and a rekeying
> operation occurred, their SSH connection would then break, because the client version
> is supposed to be idempotent.

Personally I don't think that this is hard to document and I can add a javadoc comment trying to do it to the method (I would have done so anyway, if the method didn't exist already).

src/test/java/com/jcraft/jsch/JSchTest.java
kimmerin: To `-v` that: `a >= b` can be seen as a shortcut for `a == b || a > b`. To have complete coverage over the conditions you need to implement three tests: one for `a < b`, one for `a == b` and one for `a > b`.

src/main/java/com/jcraft/jsch/Session.java
kimmerin: Added some javadoc that should explain the way this works.

norrisjeremy: I still disagree with this.
@mwiede what are your thoughts?

src/main/java/com/jcraft/jsch/Session.java
```diff
@@ -2718,15 +2719,45 @@ public void setTimeout(int timeout) throws JSchException {
...
     return Util.byte2str(V_S);
   }
+  
+  /**
+   * Returns the client version string used during an active connection.
+   * If no connection is active <code>null</code> is returned.
+   * @return The used version string or <code>null</code> if no
+   * connection is active.
+   */
+  public String getUsedClientVersion() {
```
norrisjeremy, line 2729: I personally find this extremely convoluted and would really rather we not add this method.

kimmerin: Since two members are necessary to prevent the effect you've described I though, offering a get-method to allow access to these two members should be available. But I have no special feelings about this method, so I've removed it.

src/main/java/com/jcraft/jsch/Session.java
norrisjeremy: @mwiede Any thoughts on this?
Should we just go with a simpler approach I suggested or the more complex approach implemented here?

Issue #131, opened 2022-03-09
(taken from jsch-users): Possible deadlock in KnownHosts


kimmerin:
Hi,

I've remembered that there was an issue reported on jsch-users where I've answered with a patch. I don't have the original mail in my archive on my current computer but the internet [doesn't forget](https://jsch-users.narkive.com/XfjBWvoZ/deadlock-in-knownhosts).

I've checked the current source and it seems that a part of my fix got included into the source base but not all, especially the removal of the synchronization that lead to the deadlock in the mentioned report above.

Also KnownHost contains a lot of `System.out` and `System.err.println` calls in case of errors that should be changed to logger.log-calls in my eyes.

norrisjeremy:
I'm not sure I see how this deadlock could still be occurring, at least as described in the original JSch-users thread: the `KnownHosts` class no longer contains any synchronized methods.

Are you still able to reproduce this problem with current release?

kimmerin:
> the `KnownHosts` class no longer contains any synchronized methods.

Um...

https://github.com/mwiede/jsch/blob/7e05876c65502cd0f350a5fcc75dc80e6a1da353/src/main/java/com/jcraft/jsch/KnownHosts.java#L551-L558

https://github.com/mwiede/jsch/blob/7e05876c65502cd0f350a5fcc75dc80e6a1da353/src/main/java/com/jcraft/jsch/KnownHosts.java#L582-L588

That's what I see in trunk right at this moment.

> Are you still able to reproduce this problem with current release?

The problem was reported by somebody else and I just provided a solution by looking very closely at the deadlock-message in that mail and its stacktraces. We've had occasional deadlocks as well "out there" and after this fix never heard of one again (but I might remember wrong and this was fixed by me at some other place back then).

As mentioned before in another issue I'm currently doing a lot of customization of the original sources when a new release is available (a lot of effort I try to reduce by contributing the useful stuff to this project - the logging issue being one of them). So if I do testing with SSH I always use a version without above synchronized blocks.

Above synchronized blocks are completely unnecessary. `hash()` get's only called during instantiation so that you don't have the situation that `hash()` and `isMatched` get called in parallel. The blocks are a relic of the time when there was a VM-wide hash-instance.

norrisjeremy:
Neither of those are synchronizing on the `KnownHosts` objects, they are synchronizing on the `HmacSHA1` object.

[...]

PR #134, opened 2022-03-21
Issue131

kimmerin:
Hi, as announced I've put KnownHosts under test and implemented some of the issues discussed in #131:

 - The SHA1 instance can't be `null` now. If the instantiation of the configured class fails, the JSCH SHA1-class is used as fallback. This doesn't solve things because the SSH-handshake will fail later on (but the exception there is expected and should help troubleshoot). I haven't implemented a check if the configured SHA1-class is actually SHA1 as requested by @norrisjeremy
 - I fixed a bug in remove: There is a loop over the elements where entries are taken from the vector using `getElementAt(i)`. If an element is removed, `i` isn't decreased, so one entry to be checked is skipped that might have been removed otherwise.
 - All debug-output to STDOUT/ERR have been replaced by log entries to the logger. For this I've added a default method in Logger that allows to log the causing log entry as well.
 - Created a unit test that put the whole class under test (93,4 % coverage). In order to be able to do tests, some methods were made package visible and some methods were extracted to reduce complexity.

### DIFF
--- a/src/main/java/com/jcraft/jsch/JulLogger.java
+++ b/src/main/java/com/jcraft/jsch/JulLogger.java
@@ -6,7 +6,7 @@ import java.util.logging.Logger;
 public class JulLogger implements com.jcraft.jsch.Logger {
 
   private static final java.util.logging.Logger stlogger = java.util.logging.Logger.getLogger(JSch.class.getName());
-  private java.util.logging.Logger logger;
+  private final java.util.logging.Logger logger;
   
   public JulLogger() {
     this(stlogger);
--- a/src/main/java/com/jcraft/jsch/JulLogger.java
+++ b/src/main/java/com/jcraft/jsch/JulLogger.java
@@ -1,12 +1,20 @@
 package com.jcraft.jsch;
 
 import java.util.logging.Level;
+import java.util.logging.Logger;
 
 public class JulLogger implements com.jcraft.jsch.Logger {
 
-  private static final java.util.logging.Logger logger = java.util.logging.Logger.getLogger(JSch.class.getName());
+  private static final java.util.logging.Logger stlogger = java.util.logging.Logger.getLogger(JSch.class.getName());
+  private java.util.logging.Logger logger;
+  
+  public JulLogger() {
+    this(stlogger);
+  }
 
-  public JulLogger() {}
+  JulLogger(Logger logger) {
+    this.logger = logger;
+  }
 
   @Override
   public boolean isEnabled(int level) {
@@ -15,10 +23,19 @@ public class JulLogger implements com.jcraft.jsch.Logger {
 
   @Override
   public void log(int level, String message) {
-    logger.log(getLevel(level), message);
+    log(level, message, null);
   }
 
-  private static Level getLevel(int level) {
+  @Override
+  public void log(int level, String message, Throwable cause) {
+    if (cause == null) {
+      logger.log(getLevel(level), message);
+      return;
+    }
+    logger.log(getLevel(level), message, cause);
+  }
+
+  static Level getLevel(int level) {
     switch (level) {
       case com.jcraft.jsch.Logger.DEBUG:
         return Level.FINE;
--- a/src/main/java/com/jcraft/jsch/JulLogger.java
+++ b/src/main/java/com/jcraft/jsch/JulLogger.java
@@ -5,8 +5,8 @@ import java.util.logging.Logger;
 
 public class JulLogger implements com.jcraft.jsch.Logger {
 
-  private static final java.util.logging.Logger stlogger = java.util.logging.Logger.getLogger(JSch.class.getName());
-  private final java.util.logging.Logger logger;
+  private static final Logger stlogger = Logger.getLogger(JSch.class.getName());
+  private final Logger logger;
   
   public JulLogger() {
     this(stlogger);
--- a/src/main/java/com/jcraft/jsch/KnownHosts.java
+++ b/src/main/java/com/jcraft/jsch/KnownHosts.java
@@ -39,10 +39,9 @@ import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Vector;
-import com.jcraft.jsch.jce.HMACSHA1;
 
 class KnownHosts implements HostKeyRepository{
-  private static final String _known_hosts="known_hosts";
+//  private static final String _known_hosts="known_hosts";
 
   private JSch jsch=null;
   private String known_hosts=null;
@@ -87,7 +86,7 @@ loop:
           j=fis.read();
           if(j==-1){
             if(bufl==0){ break loop; }
-            else{ break; }
+            break;
           }
           if(j==0x0d){ continue; }
           if(j==0x0a){ break; }
@@ -296,7 +295,7 @@ loop:
   public void add(HostKey hostkey, UserInfo userinfo){
     int type=hostkey.type;
     String host=hostkey.getHost();
-    byte[] key=hostkey.key;
+//    byte[] key=hostkey.key;
 
     HostKey hk=null;
     synchronized(pool){
--- a/src/main/java/com/jcraft/jsch/KnownHosts.java
+++ b/src/main/java/com/jcraft/jsch/KnownHosts.java
@@ -440,15 +440,8 @@ loop:
   }
   synchronized void sync(String foo) throws IOException {
     if(foo==null) return;
-    FileOutputStream fos=null;
-    try {
-      fos=new FileOutputStream(Util.checkTilde(foo));
+    try (FileOutputStream fos = new FileOutputStream(Util.checkTilde(foo))) {
       dump(fos);
-    }
-    finally {
-      if (fos != null) {
-        fos.close();
-      }
     }
   }
 
--- a/src/main/java/com/jcraft/jsch/KnownHosts.java
+++ b/src/main/java/com/jcraft/jsch/KnownHosts.java
@@ -50,7 +50,7 @@ class KnownHosts implements HostKeyRepository{
 
   MAC hmacsha1;
 
-  KnownHosts(JSch jsch){
+  KnownHosts(JSch jsch) {
     super();
     this.jsch=jsch;
     getHMACSHA1();
@@ -510,7 +510,7 @@ loop:
     return hosts;
   }
 
-  MAC getHMACSHA1(){
+  MAC getHMACSHA1() throws IllegalArgumentException {
     if (hmacsha1 == null){
       hmacsha1 = createHMAC(JSch.getConfig("hmac-sha1"));
     }
@@ -518,14 +518,14 @@ loop:
     return hmacsha1;
   }
   
-  MAC createHMAC(String hmacClassname) {
+  MAC createHMAC(String hmacClassname) throws IllegalArgumentException {
     try{
       Class<? extends MAC> c=Class.forName(hmacClassname).asSubclass(MAC.class);
       return c.getDeclaredConstructor().newInstance();
     }
-    catch(Exception e){ 
-      jsch.getInstanceLogger().log(Logger.ERROR, "unable to instantiate HMAC-class " + hmacClassname + ", using SHA-1 as fallback", e);
-      return new HMACSHA1(); // the default in config
+    catch(Exception e){
+      jsch.getInstanceLogger().log(Logger.ERROR, "unable to instantiate HMAC-class " + hmacClassname, e);
+      throw new IllegalArgumentException("instantiation of " + hmacClassname + " lead to an error", e);
     }
   }
 
--- a/src/main/java/com/jcraft/jsch/KnownHosts.java
+++ b/src/main/java/com/jcraft/jsch/KnownHosts.java
@@ -448,7 +448,6 @@ loop:
   private static final byte[] space={(byte)0x20};
   private static final byte[] lf=Util.str2byte("\n");
   
-  @SuppressWarnings("unused") // to keep the method's signature without warning
   void dump(OutputStream out) {
     try{
       HostKey hk;
--- a/src/main/java/com/jcraft/jsch/KnownHosts.java
+++ b/src/main/java/com/jcraft/jsch/KnownHosts.java
@@ -29,10 +29,17 @@ EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 package com.jcraft.jsch;
 
-import java.io.*;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Vector;
+import com.jcraft.jsch.jce.HMACSHA1;
 
 class KnownHosts implements HostKeyRepository{
   private static final String _known_hosts="known_hosts";
@@ -41,12 +48,12 @@ class KnownHosts implements HostKeyRepository{
   private String known_hosts=null;
   private Vector<HostKey> pool=null;
 
-  private MAC hmacsha1=null;
+  MAC hmacsha1;
 
   KnownHosts(JSch jsch){
     super();
     this.jsch=jsch;
-    this.hmacsha1 = getHMACSHA1();
+    getHMACSHA1();
     pool=new Vector<>();
   }
 
@@ -258,7 +265,8 @@ loop:
     try {
       hk = new HostKey(host, HostKey.GUESS, key);
     }
-    catch(JSchException e){  // unsupported key
+    catch(Exception e){  // unsupported key
+      jsch.getInstanceLogger().log(Logger.DEBUG, "exception while trying to read key while checking host '" + host + "'", e);
       return result;
     }
 
@@ -269,9 +277,7 @@ loop:
           if(Util.array_equals(_hk.key, key)){
             return OK;
           }
-          else{
-            result=CHANGED;
-          }
+          result=CHANGED;
         }
       }
     }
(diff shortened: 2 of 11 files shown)


src/main/java/com/jcraft/jsch/Logger.java
```diff
@@ -40,6 +43,20 @@ LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
...
+  public default void log(int level, String message, Throwable cause) {
+    if (!isEnabled(level)) {
+      return;
+    }
+    if (cause != null) {
+      StringWriter sw = new StringWriter();
+      try (PrintWriter pw = new PrintWriter(sw, true)) {
+        cause.printStackTrace(pw);
+      }
+      message += "\r\n" + sw.toString();
```
norrisjeremy, line 56: I think we should do this instead:
```
message += System.lineSeparator() + sw.toString();
```

That way it uses the correct line separator for whatever platform is in use.

norrisjeremy (changes requested):

src/main/java/com/jcraft/jsch/Logger.java
norrisjeremy: I also wonder if we should override this default method in the concrete implementations I added to utilize their specialized methods for logging exceptions?

kimmerin: I've changed the line break and added specific log-methods to framework-specific implementations. I've also added test cases for all the logging frameworks that allow logger configuration in a programmatic way, i.e. there is no test for Log4j2.

norrisjeremy:
It looks like there is a test failure with `LoggerTest.testLogWithCause`?

src/main/java/com/jcraft/jsch/Slf4jLogger.java
```diff
@@ -1,49 +1,58 @@
...
-        return logger.isInfoEnabled();
-      case com.jcraft.jsch.Logger.WARN:
-        return logger.isWarnEnabled();
-      case com.jcraft.jsch.Logger.ERROR:
-      case com.jcraft.jsch.Logger.FATAL:
-        return logger.isErrorEnabled();
-      default:
-        return logger.isTraceEnabled();
-    }
+    return logger.isEnabledForLevel(getLevel(level));
```
norrisjeremy, line 23: I think we need to revert this part: I believe the `isEnabledForLevel()` method is only available with SLFJ4 2.x, but we should keep compatibility with the SLF4J 1.x API as well.

norrisjeremy (changes requested):
Can you also add specialization in the `JplLogger` class too (under `src/main/java9`)?

src/main/java/com/jcraft/jsch/JulLogger.java
```diff
@@ -1,12 +1,20 @@
 package com.jcraft.jsch;
 
 import java.util.logging.Level;
+import java.util.logging.Logger;
 
 public class JulLogger implements com.jcraft.jsch.Logger {
 
-  private static final java.util.logging.Logger logger = java.util.logging.Logger.getLogger(JSch.class.getName());
+  private static final java.util.logging.Logger stlogger = java.util.logging.Logger.getLogger(JSch.class.getName());
+  private java.util.logging.Logger logger;
```
norrisjeremy, line 9: I think the can be a `private final` field?

src/main/java/com/jcraft/jsch/Slf4jLogger.java
```diff
@@ -1,49 +1,58 @@
...
+import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.slf4j.event.Level;
+import org.slf4j.spi.LoggingEventBuilder;
 
 public class Slf4jLogger implements com.jcraft.jsch.Logger {
 
-  private static final org.slf4j.Logger logger = LoggerFactory.getLogger(JSch.class);
+  private static final Logger stlogger = LoggerFactory.getLogger(JSch.class);
+  private Logger logger;
```
norrisjeremy, line 11: I think the can be a `private final` field?

norrisjeremy (changes requested):

src/main/java/com/jcraft/jsch/Slf4jLogger.java
kimmerin: I assumed that slf4j 2.0 is used because it's in the pom.xml

norrisjeremy: > I assumed that slf4j 2.0 is used because it's in the pom.xml

I switched the dependency to 2.x because the 1.x versions aren't modularized, so they caused warnings for `module-info.java` for JPMS support.
But I made sure that the `Slf4jLogger` was fully compatible with 1.x, allowing users to decide which version they want to use.

kimmerin: If backward compatibility to slf4j 1 is intended, the pom should be changed to use that library to make sure that the compiler always compiles against the correct methods.

norrisjeremy: > If backward compatibility to slf4j 1 is intended, the pom should be changed to use that library to make sure that the compiler always compiles against the correct methods.

Except the project doesn't compile `module-info.java` correctly if it is built with the 1.x version of the library, and then we lose JPMS support.

It doesn't seem that difficult to just leave the code as it is with the existing `switch` statement (it works with both SLFJ4 1.x & 2.x), and the way I structured the project, users can make use of either version in their own projects.

kimmerin: I'll make a break, then will do the change back to v1-compatibility

kimmerin: > Except the project doesn't compile `module-info.java` correctly if it is built with the 1.x version of the library, and then we lose JPMS support.

OK. I've just been bitten once already when compiling against a newer API while executing with an older one. It's been a while but a lecture for life ;-) In the good old times when a kilobyte weighted a kilogram, `java.lang.BigDecimal` got a new constructor `BigDecimal(long)`. While I don't expect something like that took place in Slf4J, it's still dangerous to do so.

norrisjeremy: Yep, I realize that and I wouldn't have changed the pom.xml to the 2.x version, except that I was also trying to make JSch fully modularized & JPMS compatible, and this was the only way I found to accomplish that.
I think as long as we are careful about what SLF4J methods we utilize, we can accomplish everything:
1. SLF4j 1.x compability.
2. SLF4j 2.x compability.
3. Java 9+ JPMS support.

kimmerin: Hm, I'm not that Maven-savvy but isn't it possible to define a test-only dependency. Can we use slf4j 2 for compiling and slf4j 1 for testing? With the test cases I've added any incompatibility (and be it compiler-related) should show up during test that way.

norrisjeremy: I think the only way to do that would be to completely restructure as a multi-module project, with the library as one module and the tests as a different module, but that would be alot of work.

kimmerin:
> Can you also add specialization in the `JplLogger` class too (under `src/main/java9`)?

done

src/main/java/com/jcraft/jsch/Slf4jLogger.java
kimmerin: done

src/main/java/com/jcraft/jsch/JulLogger.java
kimmerin: done

src/main/java/com/jcraft/jsch/Slf4jLogger.java
kimmerin: I've changed the implementation to only use the API of slf4j 1

src/main/java/com/jcraft/jsch/Slf4jLogger.java
```diff
@@ -1,48 +1,63 @@
...
+    switch(level) {
+        case com.jcraft.jsch.Logger.DEBUG:
+            return logger.isDebugEnabled();
+        case com.jcraft.jsch.Logger.ERROR:
+        case com.jcraft.jsch.Logger.FATAL:
+            return logger.isErrorEnabled();
+        case com.jcraft.jsch.Logger.INFO:
+            return logger.isInfoEnabled();
+        case com.jcraft.jsch.Logger.WARN:
+            return logger.isWarnEnabled();
```
norrisjeremy, line 30: The formatting for this no longer seems to follow [Google Java Style Guide's guidelines in section 4.8.4](https://google.github.io/styleguide/javaguide.html#s4.8.4-switch).
Can we return the formatting back, since I created this file following their style guide?

norrisjeremy: Also, I had purposely ordered the levels in the switch statement in order by severity (lowest to highest), can we also restore that order?

src/main/java/com/jcraft/jsch/Slf4jLogger.java
```diff
@@ -1,48 +1,63 @@
...
     }
+    return logger.isTraceEnabled();
```
norrisjeremy, line 32: I think we should put this back as a `default` case in the switch statement above, to match the switch statement in the new specialization for Throwable's below where it is in its `default` case.

src/main/java/com/jcraft/jsch/Slf4jLogger.java
```diff
@@ -1,48 +1,63 @@
...
-        logger.warn(message);
-        break;
-      case com.jcraft.jsch.Logger.ERROR:
-      case com.jcraft.jsch.Logger.FATAL:
-        logger.error(message);
+        logger.warn(message, cause);
         break;
       default:
-        logger.trace(message);
+        logger.trace(message, cause);
```
norrisjeremy, line 60: Can we order the cases in this switch statement to match original, which was by severity?

norrisjeremy (changes requested):

src/main/java/com/jcraft/jsch/KnownHosts.java
```diff
@@ -425,52 +440,65 @@ void sync() throws IOException {
...
+    FileOutputStream fos=null;
+    try {
+      fos=new FileOutputStream(Util.checkTilde(foo));
+      dump(fos);
+    }
+    finally {
+      if (fos != null) {
+        fos.close();
+      }
+    }
```
norrisjeremy, line 452: I wonder if we should change this to a try-with-resources statement?

src/main/java/com/jcraft/jsch/KnownHosts.java
```diff
@@ -425,52 +440,65 @@ void sync() throws IOException {
...
   }
 
   private static final byte[] space={(byte)0x20};
-  private static final byte[] cr=Util.str2byte("\n");
+  private static final byte[] lf=Util.str2byte("\n");
+  
+  @SuppressWarnings("unused") // to keep the method's signature without warning
```
norrisjeremy, line 458: Curious as to why this is needed since the `out` argument appears to be used in the call to `dumpHostKey()`?

src/main/java/com/jcraft/jsch/Slf4jLogger.java
```diff
@@ -27,22 +41,30 @@ public boolean isEnabled(int level) {
 
   @Override
   public void log(int level, String message) {
+    log (level, message, null);
```
norrisjeremy, line 44: Minor nit: there appears to be an extraneous space character between `log` and `(`.

src/main/java/com/jcraft/jsch/JulLogger.java
```diff
@@ -1,12 +1,20 @@
 package com.jcraft.jsch;
 
 import java.util.logging.Level;
+import java.util.logging.Logger;
 
 public class JulLogger implements com.jcraft.jsch.Logger {
 
-  private static final java.util.logging.Logger logger = java.util.logging.Logger.getLogger(JSch.class.getName());
+  private static final java.util.logging.Logger stlogger = java.util.logging.Logger.getLogger(JSch.class.getName());
+  private final java.util.logging.Logger logger;
```
norrisjeremy, line 9: I think since you've added an `import java.util.logging.Logger;` statement, there is no longer a need to specify the fully qualified package for these?

norrisjeremy: Although I wonder if it would be better to drop the `import java.util.logging.Logger` statement instead and just always fully qualify the package everywhere? That would make it slightly easier to reason about which `Logger` class is being used.

src/main/java/com/jcraft/jsch/Slf4jLogger.java
```diff
@@ -1,12 +1,26 @@
 package com.jcraft.jsch;
 
+import org.slf4j.Logger;
```
norrisjeremy, line 3: I wonder if it would be better to drop this import statement and just always fully qualify the package for the `Logger` references below so that it is easier to reason about which type of `Logger` is being referred to?

kimmerin: I think, the class is simple enough that one can be done by import and the other fully qualified. IMHO the fully qualified way is less readable.

norrisjeremy: Well, it does lead to a bit of inconsistencies between various implementations, as some of the now have an import statement for the logging framework's `Logger` class and some don't.

We should make it consistent between them all whether they do or don't import it.

src/main/java/com/jcraft/jsch/Log4j2Logger.java
```diff
@@ -19,6 +19,15 @@ public void log(int level, String message) {
...
 
+  @Override
+  public void log(int level, String message, Throwable cause) {
+    if (cause != null) {
+      logger.log(getLevel(level), message);
+      return;
+    }
+    logger.log(getLevel(level), message, cause);
+  }
+
```
norrisjeremy, line 30: Since we added an `import` statement for the `Logger` class for the other implementations, we should probably do the same thing for this class now.

src/main/java/com/jcraft/jsch/Slf4jLogger.java
kimmerin: I've checked the loggers and haven't found any inconsistencies

src/main/java/com/jcraft/jsch/KnownHosts.java
norrisjeremy: I'm not sure I understand why this warning suppression is required?

kimmerin: `out` is used but no `IOException` is thrown. You can't remove it from the method signature without potentially breaking things, so my solution was to suppress the warning and keep the method signature as it was before.

src/main/java/com/jcraft/jsch/Log4j2Logger.java
kimmerin: I think I've changed that already without seeing your comment, haven't I?

src/main/java/com/jcraft/jsch/KnownHosts.java
norrisjeremy: Why not just remove the `throws IOException` then from the method signature?
It’s a package private method, so it’s not part of the public facing API.

kimmerin: I've checked and you're right. Except the testcase I've created this didn't break anything else, so I've changed it

src/main/java/com/jcraft/jsch/KnownHosts.java
```diff
@@ -490,18 +510,24 @@ private String deleteSubString(String hosts, String host){
...
+  MAC createHMAC(String hmacClassname) {
+    try{
+      Class<? extends MAC> c=Class.forName(hmacClassname).asSubclass(MAC.class);
+      return c.getDeclaredConstructor().newInstance();
+    }
+    catch(Exception e){ 
+      jsch.getInstanceLogger().log(Logger.ERROR, "unable to instantiate HMAC-class " + hmacClassname + ", using SHA-1 as fallback", e);
+      return new HMACSHA1(); // the default in config
+    }
+  }
```
norrisjeremy, line 530: So I've been thinking about this some more: I think instead of falling back to the `com.jcraft.jsch.jce.HMACSHA1` implementation, I think we should simply throw a `JSchException` if creating the `hmacClassname` fails.
I think this will work because:

1. The `KnownHosts` class is package private since the 0.2.1 release, so we have some API flexibility.
2. All the locations in the `JSch` and `Session` classes where we instantiate a `KnownHosts` instance all appear to be in methods that already allow throwing a `JSchException`.

I also think this makes more sense from a user perspective, since it would immediately inform a user that the class they configured as `hmac-sha1` was not usable, instead of hiding it away in a way that they may not realize (especially if they don't have a logger setup).

norrisjeremy (changes requested):

src/main/java/com/jcraft/jsch/KnownHosts.java
kimmerin: I can do the change but a misconfigured hmac-class should lead to execption during SSH-handshakes as you've pointed out in the thread of the issue itself.

norrisjeremy: Yes, but I don't think we should rely on that behavior.

Also, whilst the `KnownHosts` class isn't public, it is a `HostKeyRepository` implementation, which a user can obtain access to via the `JSch.getHostKeyRepository()` or `Session.getHostKeyRepository()`, which they could then utilize standalone for their own purposes, before creating a session that would trigger the SSH-handshake issue.
Therefore, I think it's best that we not hide-away the potential `hmac-sha1` misconfiguration issue inside the `KnownHosts` class, but rather fail immediately when it is encountered.

kimmerin: > All the locations in the JSch and Session classes where we instantiate a KnownHosts instance all appear to be in methods that already allow throwing a JSchException.

That's actually not the case:

<pre>
  public HostKeyRepository getHostKeyRepository(){ 
    if(known_hosts==null) known_hosts=new KnownHosts(this);
    return known_hosts; 
  }
</pre>

I can put it into a RuntimeException and throw that instead

norrisjeremy: Oops, you are correct. Throwing a RuntimeException is probably the best choice, thanks!

src/main/java/com/jcraft/jsch/KnownHosts.java
```diff
@@ -29,10 +29,17 @@ LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
...
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Vector;
+import com.jcraft.jsch.jce.HMACSHA1;
```
norrisjeremy, line 42: I believe this import can be removed?

norrisjeremy (changes requested):

src/main/java/com/jcraft/jsch/KnownHosts.java
kimmerin: It can. Missed that because there were other warnings of unused stuff. I've commented out the other lines containing unused stuff as well (not removed, maybe there is a reason why it's currently unused)

src/main/java/com/jcraft/jsch/KnownHosts.java
```diff
@@ -29,24 +29,30 @@ LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Vector;
 
 class KnownHosts implements HostKeyRepository{
-  private static final String _known_hosts="known_hosts";
+//  private static final String _known_hosts="known_hosts";
```
norrisjeremy, line 44: If this isn't used, then I would say to remove it instead of commenting it out.

src/main/java/com/jcraft/jsch/KnownHosts.java
```diff
@@ -290,7 +295,7 @@ public int check(String host, byte[] key){
   public void add(HostKey hostkey, UserInfo userinfo){
     int type=hostkey.type;
     String host=hostkey.getHost();
-    byte[] key=hostkey.key;
+//    byte[] key=hostkey.key;
```
norrisjeremy, line 298: If this isn't used, then I would say to remove it instead of commenting it out.

norrisjeremy (changes requested):

src/main/java/com/jcraft/jsch/KnownHosts.java
kimmerin: I can delete it (as with the others) but my usual workflow is commenting out first and if nobody complains within some given time, I delete it in a later step.

src/main/java/com/jcraft/jsch/KnownHosts.java
kimmerin: `key` is used in a code block that currently is commented out a couple of lines later. So the person who commented out that block simply forgot to do it with this particular line. That's why I'd keep it there until it's decided what should happen with the larger code block. If it's "reactivated", the line can be brought back to life as well and if it's removed that line has to go, too.

src/main/java/com/jcraft/jsch/KnownHosts.java
norrisjeremy: I think we can just safely remove it since it is a private field.

src/main/java/com/jcraft/jsch/KnownHosts.java
norrisjeremy: Ok, you are correct on this one: it is best to just comment it out for now as you have done.

norrisjeremy (approved):
LGTM.
@mwiede I think we should do a squash merge of this so that's all contained with a single commit.

PR #141, opened 2022-04-25
Using correct line ending CRLF instead of \n (LF)

guidowalter:
According to https://datatracker.ietf.org/doc/html/rfc4253#section-4.2 line must end by using CRLF not \n. 
Missing that creates connection issues e.g. https://docs.microsoft.com/en-us/answers/questions/713024/connection-to-azure-sftp-doesnt-work-using-jsch.html

### DIFF
--- a/src/main/java/com/jcraft/jsch/Session.java
+++ b/src/main/java/com/jcraft/jsch/Session.java
@@ -263,7 +263,7 @@ public class Session{
         // Some Cisco devices will miss to read '\n' if it is sent separately.
         byte[] foo=new byte[V_C.length+2];
         System.arraycopy(V_C, 0, foo, 0, V_C.length);
-        foo[foo.length-2]=(byte)'\r';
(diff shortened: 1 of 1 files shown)


src/main/java/com/jcraft/jsch/Session.java
```diff
@@ -263,7 +263,7 @@ public void connect(int connectTimeout) throws JSchException{
         // Some Cisco devices will miss to read '\n' if it is sent separately.
         byte[] foo=new byte[V_C.length+2];
         System.arraycopy(V_C, 0, foo, 0, V_C.length);
-        foo[foo.length-2]=(byte)'\r';
+        foo[foo.length-2]=(byte)0x0D;
```
norrisjeremy, line 266: I don't understand this: `'\r'` is `0x0D`...
```
jshell> var test = (byte)'\r' == (byte)0x0D;
test ==> true
```

norrisjeremy (changes requested):

PR #151, opened 2022-05-08
setup jdk for code-ql analysis

mwiede:


norrisjeremy:
Is this new workflow cancelling the maven build workflows because it fails?

mwiede:
> Is this new workflow cancelling the maven build workflows because it fails?

it should not. I can change the setting if it would block it.

PR #157, opened 2022-06-04
Update link to bcrypt

turbanoff:
Old link doesn't work
изображение

### DIFF
--- a/opensshFormat.md
+++ b/opensshFormat.md
@@ -6,4 +6,4 @@
 <https://bugs.eclipse.org/bugs/show_bug.cgi?id=541703>
 
 * implementation of bcrypt_pbkdf:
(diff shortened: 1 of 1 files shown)


opensshFormat.md
```diff
@@ -6,4 +6,4 @@
 <https://bugs.eclipse.org/bugs/show_bug.cgi?id=541703>
 
 * implementation of bcrypt_pbkdf:
-<https://github.com/kruton/jbcrypt/blob/master/jbcrypt/src/main/java/org/mindrot/jbcrypt/BCrypt.java>
\ No newline at end of file
+<https://github.com/kruton/jbcrypt/blob/master/src/main/java/org/mindrot/jbcrypt/BCrypt.java>
```
norrisjeremy, line 9: I wonder if we should instead just link to the copy that is maintained directly in this repo instead?

PR #175, opened 2022-07-14
Improve performance.

wisdt:
Improve performance for multiple threads.

### DIFF
--- a/src/main/java/com/jcraft/jsch/Channel.java
+++ b/src/main/java/com/jcraft/jsch/Channel.java
@@ -30,7 +30,8 @@ EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 package com.jcraft.jsch;
 
 import java.io.*;
-import java.util.Vector;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
 
 public abstract class Channel{
 
@@ -43,8 +44,8 @@ public abstract class Channel{
   static final int SSH_OPEN_UNKNOWN_CHANNEL_TYPE=           3;
   static final int SSH_OPEN_RESOURCE_SHORTAGE=              4;
 
-  static int index=0; 
-  private static Vector<Channel> pool=new Vector<>();
+  private static AtomicInteger index = new AtomicInteger(0);
+  private static ConcurrentHashMap<Integer, Channel> pool = new ConcurrentHashMap<>(16, 0.75f, 1);
   static Channel getChannel(String type, Session session){
     Channel ret = null;
     if(type.equals("session")){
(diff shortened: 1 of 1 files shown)


src/main/java/com/jcraft/jsch/Channel.java
```diff
@@ -128,10 +125,8 @@ static void del(Channel c){
   int notifyme=0; 
 
   Channel(){
-    synchronized(pool){
-      id=index++;
-      pool.addElement(this);
-    }
+    id=index.incrementAndGet();
```
norrisjeremy, line 128: Shouldn't this be `id=index.getAndIncrement()` in order to maintain the same semantics as `id=index++`?

src/main/java/com/jcraft/jsch/Channel.java
```diff
@@ -43,8 +44,8 @@ LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
   static final int SSH_OPEN_UNKNOWN_CHANNEL_TYPE=           3;
   static final int SSH_OPEN_RESOURCE_SHORTAGE=              4;
 
-  static int index=0; 
-  private static Vector<Channel> pool=new Vector<>();
+  private static AtomicInteger index = new AtomicInteger(0);
+  private static ConcurrentHashMap<Integer, Channel> pool = new ConcurrentHashMap<>(16, 0.75f, 1);
```
norrisjeremy, line 48: 1. This should be `ConcurrentMap<Integer, Channel> pool` instead of `ConcurrentHashMap<Integer, Channel> pool`.
2. Can you provide some reasoning for why these particular values were chosen for the ConcurrentHashMap constructor?

norrisjeremy (changes requested):
My gut feeling is that the act of creating and disposing of channels does not occur at a high frequency in practice, meaning that making these changes purely for performance reasons probably doesn't result in any meaningful benefit for live applications.

And it also makes me a bit nervous to touch stuff like this since it could lead to subtle regressions that we don't immediately detect (since we lack a good test suite to run changes like this through). Therefore I'm not especially inclined to be in favor of landing this change.

That said: are you able to provide any benchmarks to demonstrate quantifiable improvements that are brought by this change?

wisdt:
Could you delete this pull-request please. I created it by mistake from a personal account. Also need to write a more detailed description and collect information of performance testing in order to submit a more detailed and complete pull-request.

src/main/java/com/jcraft/jsch/Channel.java
```diff
@@ -550,18 +545,11 @@ public boolean isClosed(){
...
-        try{
-          Channel c=pool.elementAt(i);
-          if(c.session==session){
-            channels[count++]=c;
-          }
-        }
-        catch(Exception e){
-        }
-      } 
+    channels=new Channel[pool.size()];
```
norrisjeremy, line 548: This looks problematic, in that the size of `pool` could increase after the call to `pool.size()` here and before the iteration over `pool.values()` below, potentially leading to an `ArrayIndexOutOfBoundsException` being thrown.

I think that instead of accumulating all the channels with a matching session into an array, and then iterating over this array to call `disconnect()`, we could simply hoist the call to `disconnect()` directly in the first iteration (that checks for matching sessions).

I believe the only reason it was done this way (storing matching sessions in an array and then iterating over said array) was so that the calls to `disconnect()` didn't happen while `pool` was locked.

Since with this change you are no longer locking the `pool` object, it should be safe to directly call `disconnect()` while iterating through `pool`.

norrisjeremy (commented):
FYI, I realize you closed this for now, but I wanted to leave an additional comment.

src/main/java/com/jcraft/jsch/Channel.java
wisdt: I got it, I will improve it. Please delete my misscreated PullRequest.

PR #241, opened 2022-11-16
0.2.5 changes

norrisjeremy:
* Explictly free resources in `Compression` implementations.


norrisjeremy:
> Hmm, I think I missed one `if` condition, I'll update this later today or tomorrow.

Ok, I think this is corrected now.

PR #247, opened 2022-11-29
activate sourcecode formatting

mwiede:
This will use formatting rules of `eclipse-java-google-style.xml` to format the source code in maven `process-sources` phase.
There should be only whitespace changes in java files.

Checkstyle Plugin is added, but not used by default, since current code would have 2640 violations against `Google-Style`.
```
$ mvn validate -Pcheckstyle
...
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  23.848 s
[INFO] Finished at: 2022-11-29T10:34:30Z
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.2.0:check (validate) on project jsch: You have 2640 Checkstyle violations. -> [Help 1]
```
Cleaning these up is an additional effort.


norrisjeremy:
Have you considered using [fmt-maven-plugin](https://github.com/spotify/fmt-maven-plugin)?
It directly utilizes Google [google-jave-format](https://github.com/google/google-java-format) library to process sources.
We typically add it's check goal like this to our projects so that non-conforming code causes a build failure during the `verify` phase:
```
      <plugin>
        <groupId>com.spotify.fmt</groupId>
        <artifactId>fmt-maven-plugin</artifactId>
        <version>2.19</version>
        <executions>
          <execution>
            <goals>
              <goal>check</goal>
            </goals>
            <phase>verify</phase>
          </execution>
        </executions>
      </plugin>
```

And to directly use it's format goal to reformat sources, you can manually execute `mvn fmt:format`.
It may require some additional tweaks to the plugin configuration to pick-up all the non-standard source directories where we put the > Java 8 source code.

mwiede:
i have not compared fmt-maven-plugin, but as far as i can see formatter-maven-plugin offers the same. you can both format source code as well as verify it.

norrisjeremy:
It appears `formatter-maven-plugin` also has a `validate` goal: is there a reason to to use `maven-checkstyle-plugin` instead of that?

Also, do we need to configure these plugins to also handle the > Java 8 code directories (`src/main/java9`, `src/main/java11`, etc.)?

src/main/java/com/jcraft/jsch/AgentConnector.java
```diff
@@ -1,36 +1,36 @@
 /* -*-mode:java; c-basic-offset:2; indent-tabs-mode:nil -*- */
```
norrisjeremy, line 1: I think we may want to also remove all these editor hints at the top of all the files too?

pom.xml
```diff
@@ -618,5 +618,60 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
+            <id>format</id>
+            <activation>
+                <activeByDefault>true</activeByDefault>
```
norrisjeremy, line 624: Are we sure we want reformatting to be `activeByDefault`?
Or should it be that `maven-checkstyle-plugin` is `activeByDefault`?

mwiede: if we run formatting by default, there is not need to run checkstyle. It's just optional to make it available, but at the moment, the result of reformatting does not remove any of the checkstyle violations.

src/main/java/com/jcraft/jsch/AgentConnector.java
mwiede: yes makes sense

pom.xml
```diff
@@ -618,5 +618,71 @@
...
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-checkstyle-plugin</artifactId>
+                        <version>3.2.0</version>
+                        <configuration>
+                            <encoding>UTF-8</encoding>
+                            <failOnViolation>true</failOnViolation>
+                            <logViolationsToConsole>true</logViolationsToConsole>
+                            <violationSeverity>warning</violationSeverity>
+                            <configLocation>google_checks.xml</configLocation>
+                        </configuration>
```
norrisjeremy, line 673: Does `maven-checkstyle-plugin` need configuration for the non-standard source directories as well?

pom.xml
norrisjeremy: Does checkstyle report violations even after formatter runs and adjusts everything though?

norrisjeremy: Ok, I see now that checkstyle logs a huge number of warnings and the build fails when it is activated, so I see now why you didn't activate it by default.

pom.xml
```diff
@@ -618,5 +646,64 @@
...
             </build>
         </profile>
+        <profile>
+            <id>format</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>net.revelc.code.formatter</groupId>
+                        <artifactId>formatter-maven-plugin</artifactId>
+                        <version>2.21.0</version>
```
norrisjeremy, line 656: I suspect the `<version>` doesn't need to be specified here, as it should pick it up from earlier?

norrisjeremy (commented):
Do we know what's causing the test failures during CI builds?

mwiede:
> Do we know what's causing the test failures during CI builds?

I looked at it. The problem seems to be the docker bind mount of the temporary directory. We need to move the temp directory to the current workspace.

mwiede:
> > Do we know what's causing the test failures during CI builds?
> 
> I looked at it. The problem seems to be the docker bind mount of the temporary directory. We need to move the temp directory to the current workspace.

that was a problem I had within Github Codespaces environment. The other problem was not resolved by downgrading the runner image to ubuntu 20.04

pom.xml
mwiede: removed it

src/main/java/com/jcraft/jsch/AgentConnector.java
norrisjeremy: Did we want to handle removal of these editor hints in this PR or do that separately

mwiede: I will add it to this one

pom.xml
```diff
@@ -509,6 +509,26 @@
                 <artifactId>maven-dependency-plugin</artifactId>
                 <version>3.5.0</version>
             </plugin>
+            <plugin>
+                <groupId>net.revelc.code.formatter</groupId>
+                <artifactId>formatter-maven-plugin</artifactId>
+                <version>2.21.0</version>
```
norrisjeremy, line 515: 2.22.0 is the newest version.

pom.xml
```diff
@@ -612,5 +632,84 @@
...
+            </build>
+        </profile>
+        <profile>
+            <id>checkstyle</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-checkstyle-plugin</artifactId>
+                        <version>3.2.0</version>
```
norrisjeremy, line 682: 3.2.1 is the newest version.

pom.xml
```diff
@@ -612,5 +632,84 @@
...
+                                    <goal>format</goal>
+                                </goals>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+        <profile>
+            <activation>
```
norrisjeremy, line 655: We should add an explicit `<id>` for this profile.

norrisjeremy:
I think we may want to add `-Dformatter.skip=true` in `.github/actions/maven.yml` for the `Test with Maven` step.
I believe that will prevent Github CI from executing the formatter validation check when it executes the unit & integration tests (since it already would have done this during the prior `Build with Maven` step).

norrisjeremy:
> I think we may want to add `-Dformatter.skip=true` in `.github/workflows/maven.yml` for the `Test with Maven` step.
> I believe that will prevent Github CI from executing the formatter validation check when it executes the unit & integration tests (since it already would have done this during the prior `Build with Maven` step).

Disregard, I missed that formatter:validate is tied to verify phase, so it has to happen during the `Test with Maven` step.

pom.xml
mwiede: updated

pom.xml
mwiede: updated

pom.xml
mwiede: 👍

pom.xml
```diff
@@ -612,5 +632,85 @@
...
+                                    <goal>format</goal>
+                                </goals>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+        <profile>
+			<id>formatvalidation</id>
```
norrisjeremy, line 655: I think the indentation is off here?

norrisjeremy (approved):
I think we should do a squash merge of this so it's a single commit.

Issue #248, opened 2022-11-30
Add OSGI metadata to MANIFEST


zspitzer:
Would it be possible to add OSGI metadata to the Manifest?

I work on the Lucee project https://github.com/lucee/Lucee and we are currently using the old jsch lib and would like to switch to using this actively maintained fork https://luceeserver.atlassian.net/browse/LDEV-4299 

We've been wrapping up the old lib as an OSGI bundle (https://github.com/lucee/osgi-bundle-jsch), but it would be great if we could just use this one directly via maven

https://github.com/srikanth-lingala/zip4j/issues/89

mwiede:
@zspitzer can you please try 0.2.5 release and verify that it suits your needs? Thanks

zspitzer:
@mwiede danke! all working as expected

PR #249, opened 2022-11-30
generate osgi bundle manifest data for jar #248

mwiede:
changes according to https://felix.apache.org/documentation/subprojects/apache-felix-maven-bundle-plugin-bnd.html#_adding_osgi_metadata_to_existing_projects_without_changing_the_packaging_type

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -531,14 +531,19 @@
             <plugin>
                 <groupId>org.apache.felix</groupId>
                 <artifactId>maven-bundle-plugin</artifactId>
-                <version>5.1.1</version>
+                <version>5.1.8</version>
+                <configuration>
+                    <instructions>
+                        <Export-Package>com.jcraft.jsch.*</Export-Package>
+                    </instructions>
+                </configuration>
                 <executions>
                     <execution>
-                    <id>bundle-manifest</id>
-                    <phase>process-classes</phase>
-                    <goals>
-                        <goal>manifest</goal>
-                    </goals>
+                        <id>bundle-manifest</id>
+                        <phase>process-classes</phase>
+                        <goals>
+                            <goal>manifest</goal>
+                        </goals>
                     </execution>
                 </executions>
             </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -355,6 +355,25 @@
                 </executions>
             </plugin>
             <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <version>5.1.8</version>
+                <configuration>
+                    <instructions>
+                        <Export-Package>com.jcraft.jsch</Export-Package>
+                    </instructions>
+                </configuration>
+                <executions>
+                    <execution>
+                        <id>bundle-manifest</id>
+                        <phase>process-classes</phase>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -527,6 +528,20 @@
                 <artifactId>maven-dependency-plugin</artifactId>
                 <version>3.3.0</version>
             </plugin>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <version>5.1.1</version>
```
norrisjeremy, line 534: I think `5.1.8` is the most recent version of this plugin? Any reason to use an older version?

Also, do we need to add an `<extensions>true</extensions>` statement here? I've seen it in other projects, not sure exactly what it does though?

pom.xml
```diff
@@ -527,6 +528,20 @@
...
+                <executions>
+                    <execution>
+                    <id>bundle-manifest</id>
+                    <phase>process-classes</phase>
+                    <goals>
+                        <goal>manifest</goal>
+                    </goals>
+                    </execution>
```
norrisjeremy, line 542: I think we're missing some indentation here between the opening and closing `execution` tags?

pom.xml
```diff
@@ -527,6 +528,20 @@
                 <artifactId>maven-dependency-plugin</artifactId>
                 <version>3.3.0</version>
             </plugin>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <version>5.1.1</version>
+                <executions>
```
norrisjeremy, line 535: Should we add something like this to lock-in that the only public-use package is `com.jcraft.jsch` (similar to what I did in `module-info.java` for JPMS)?
```
                <configuration>
                    <instructions>
                        <Export-Package>com.jcraft.jsch</Export-Package>
                    </instructions>
                </configuration>
```

pom.xml
mwiede: fixed

pom.xml
mwiede: I have compared the output in the manifest with and without the `extensions` tag and there are no differences.

pom.xml
```diff
@@ -527,6 +528,25 @@
                 <artifactId>maven-dependency-plugin</artifactId>
                 <version>3.3.0</version>
             </plugin>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <version>5.1.8</version>
+                <configuration>
+                    <instructions>
+                        <Export-Package>com.jcraft.jsch.*</Export-Package>
```
norrisjeremy, line 537: I think we just want to export the `com.jcraft.jsch` package?

norrisjeremy (commented):
I was keeping the order of the maven plugins approximately in order of the phases in which their specified goals were executed.
Therefore, I would put `maven-bundle-plugin` after `maven-compiler-plugin` and before `scala-maven-plugin` since the `process-classes` phase happens after the `compile` phase but before the `test-compile` phase.

norrisjeremy (approved):

PR #259, opened 2023-01-02
Enhancement to present KeyPair.getKeyTypeString() method

Alex-Vol-Amz:
The getSshName() method was available for most KeyPair implementations but inconsistently and not as part of the KeyPair class interface. This makes the method part of the KeyPair class interface and makes each implementation provide the correct value to the caller. I found this to be useful when importing keys from unknown files and expecting the KeyPair class to provide a hint as to the loaded key type in a way that can be used by code logic. I prefer this to doing class type inspection as it is much easier to get it right expecially for the ECDSA KeyPair which can have different SshName based on the key size.

### DIFF
--- a/src/main/java/com/jcraft/jsch/KeyPair.java
+++ b/src/main/java/com/jcraft/jsch/KeyPair.java
@@ -177,6 +177,15 @@ public abstract class KeyPair{
   public abstract int getKeyType();
 
   /**
+   * Wrapper to provide the String representation of {@code #getKeyTypeName()}.
+   *
+   * @return the standard SSH key type string
+   */
+  public String getKeyTypeString() {
+    return Util.byte2str(getKeyTypeName());
+  }
+
+  /**
    * Returns the blob of the public key.
    * @return blob of the public key
    */


norrisjeremy:
There's already a `getKeyType()` method: is there a reason that isn't sufficient?

Alex-Vol-Amz:
The key type only gives a coarse level type. ECDSA keys for example have an SSH name that includes the key size encoded in them. I find this useful to have in my use cases. Since it is already available I see no hard in the added public method. It does not cause any backwards compatibility issues as far as I know and it is a convenient property to have.

norrisjeremy:
If we're going to promote this as a public API method, I'd like to name it something better than `getSshName()`.
I originally added it only to KeyPairEdDSA as an internal method to allow easy abstraction between Ed25519 & Ed448 and didn't consider it for use as a public API method.

How about naming it `getKeyTypeString()`? And then we can just add it directly to `KeyPair.java` as a wrapper around `getKeyTypeName()`, avoiding any changes to implementation classes:

```
public String getKeyTypeString() {
  return Util.byte2str(getKeyTypeName());
}
```

Alex-Vol-Amz:
I appreciate your suggestion. I will make the change to the new name and see if the getKeyTypeName() wrapper works.

I will also add Javadoc to the new method which I did not in the first revision.

Issue #262, opened 2023-01-05
Support for optional fallback to ssh-rsa public key algorithm in case target server doesn't support "server-sig-algs" extension


jgolda:
Hello, 

I'm using your implemetation of JSch in a project done at the company i work for. After upgrading from 1.5.5 to 1.7.2 i faced problems with an SFTP server, which doesn't support server-sig-algs extension and most likely doesn't support rsa-sha2-256 or rsa-sha2-512 as well. Most likely it's legacy AF and should be updated or replaced, but the company's client still uses it and it might not be possible to persuade him to move to anything else.
Unfortunatelly this server also fails to correctly follow the public key authentication flow as described in RFC 4252 section 7. It responds to initial SSH_MSG_USERAUTH_REQUEST containing one of the above algorithms with SSH_MSG_USERAUTH_PK_OK, but then it returns SSH_MSG_USERAUTH_FAILURE in response to the second one, resulting with authentication failure. 

It sounds quite incorrect given the requirements specified in RFC. On the other hand, this behaviour is also described in RFC 8332, section 3.3:
"Implementation experience has shown that there are servers that apply
authentication penalties to clients attempting public key algorithms
that the SSH server does not support.
(...)
When authenticating with an RSA key against a server that does not
implement the "server-sig-algs" extension, clients MAY default to an
"ssh-rsa" signature to avoid authentication penalties."

I need to handle it this way or another so I'm thinking about contributing this feature to JSch if you consider that it's worth introducing to the project. It sounds quite doable. If not, i'll have to work around it in some other way in the application space.
Please let me know of your thoughts on this.

Best regards,
Jacek

norrisjeremy:
Hi @jgolda,

Does simply following the instructions in the [FAQ](https://github.com/mwiede/jsch#why-do-ssh-rsa-type-keys-not-work-with-this-jsch-fork-and-my-server) to manually enable ssh-rsa not work?

Thanks,
Jeremy

jgolda:
I'm afraid it won't. The problem (or sort of) lies in com.jcraft.jsch.UserAuthPublicKey class. 
1. Around line 200 Jsch sends the initial request with value 0 (indicating, that it's only checking if the server supports the public key algorithm rsa-sha2-512). 
2. The server sends SSH_MSG_USERAUTH_PK_OK in response around line 210
3. Around line 286 JSch sends the actual authentication request with generated signature 
[...]

Issue #71, opened 2021-09-08
Support PuTTY's PPK version 3 private key format


norrisjeremy:
It appears that starting with release 0.75, PuTTY added a new PPK version 3 private key format.
It is documented here: https://tartarus.org/~simon/putty-snapshots/htmldoc/AppendixC.html.

PR #264, opened 2023-01-05
0.2.7 changes

norrisjeremy:
- Use slfj4-test instead of logback for unit & integration tests.
- Fix exception logging in Log4j2Logger.
- Add unit tests for Log4j2Logger.
- Tweak JulLogger & JulLoggerTest to better match other loggers & tests.
- Add unit tests for Log4j2Logger's getLevel method.
- Misc cleanup of logger tests.
- Misc cleanup of LoggerTest.
- Remove unneeded logging.
- Update dependencies.
- #265 change buffer_margin computation to be dynamic based upon the MAC to allow connections that advertise small maximum packet sizes.
- #266 fix PuTTY key parsing to work with unix line endings.
- Convert from Hashtable & Vector to HashMap & ArrayList.
- Add support for ECDSA & EdDSA type PuTTY keys.
- #71 add support for PuTTY version 3 format keys.
  - Encrypted PuTTY version 3 format keys requires [Bouncy Castle](https://www.bouncycastle.org/java.html) (bcprov-jdk18on).
- Add coverage reports for integration tests.
- Archive jars.
- Unroll passphrase key derivation for PuTTY version 2 keys.
- Add logging when exceptions are swallowed.
- Test against char representation instead of int to improve readability.
- Eliminate KeyPairDeferred and instead change handling of OpenSSH V1 type keys to be more like other KeyPair types.
- Be more vigilant about clearing private key data.
- Improve PKCS8 key handling and add support for PKCS5 2.1 encryption.
- Add support for ECDSA type PKCS8 keys.
- Remove unneeded null checks.
- Add support for SCrypt type KDF for PKCS8 keys.
- Minor tweaks of PKCS8 handling.
  - PKCS8 keys using SCrypt requires [Bouncy Castle](https://www.bouncycastle.org/java.html) (bcprov-jdk18on).
- Make some internal methods static.
- Mark Argon2 type and versions as final.
- Wrap NoClassDefFoundErrors when Bouncy Castle is unavailable.
- Add a common bzero routine.
- Mark PBKDF interface and implementation as deprecated: PBKDF2 should be used instead.
- Add support for EdDSA type PKCS8 keys.
  - EdDSA type PKCS8 keys requires [Bouncy Castle](https://www.bouncycastle.org/java.html) (bcprov-jdk18on).
- Add a parent KDF type for all the various KDF implementations.
- Minor syntax cleanup to scala sources.
- Misc cleanup to ASN1 parsing.
- Fix warnings in scala sources.
- Convert JZlib tests from Scala to Java.
- Attempt to authenticate using other signature algorithms supported by the same public key.
  - Allow the new behavior to potentially try other algorithms to be disabled via a new config option `try_additional_pubkey_algorithms`.
  - This should help address both #262 & #271.
  - Additionally, add a new config option `enable_pubkey_auth_query` to allow skipping auth queries and proceed directly to attempting full SSH_MSG_USERAUTH_REQUEST's.
- Use standard Base64 instead of commons-codec version.
- Add an `enable_auth_none` config option.
  - The `enable_auth_none` config option controls whether an initial authentication request for the method `none` is sent to detect all the supported authentication methods available on the server.


norrisjeremy:
FYI, working on an additional change this morning, should push it to this branch in a bit hopefully.

PR #279, opened 2023-02-14
0.2.8 changes

norrisjeremy:
- Update dependencies.
- Add additional enforcer rules.
- Remove unneeded junixsocket-native-common dependency.
- Generate CycloneDX SBOM.
- Allow manual dispatch of CI builds.
- Fix CI builds with ubuntu-latest by deleting conflicting users & groups from container.
- Capture maven version info during CI builds.
- Only fork for error-prone where it's needed.
- Define plugin verions and configuration in default build instead within profiles.
  - This will help dependabot, as I don't believe it will generate PRs for items that have their versions' only defined within a profile.
- Drop compiler-annotations dependency by disabling classfile warnings.
- Correct license name for JZlib so that it isn't misidentified in SBOM.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -96,6 +96,7 @@
             <groupId>com.kohlschutter</groupId>
             <artifactId>compiler-annotations</artifactId>
             <version>1.5.2</version>
+            <scope>provided</scope>
             <optional>true</optional>
         </dependency>
         <dependency>
@@ -479,6 +480,21 @@
                         <id>attach-javadocs</id>
                         <goals>
                             <goal>jar</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.cyclonedx</groupId>
+                <artifactId>cyclonedx-maven-plugin</artifactId>
+                <version>2.7.4</version>
+                <configuration>
+                    <includeProvidedScope>false</includeProvidedScope>
+                </configuration>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>makeBom</goal>
                         </goals>
                     </execution>
                 </executions>
--- a/pom.xml
+++ b/pom.xml
@@ -63,7 +63,7 @@
             <url>https://github.com/mwiede/jsch/blob/master/LICENSE.txt</url>
         </license>
         <license>
-            <name>BSD</name>
+            <name>Revised BSD</name>
             <url>https://github.com/mwiede/jsch/blob/master/LICENSE.JZlib.txt</url>
         </license>
         <license>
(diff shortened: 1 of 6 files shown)


src/test/resources/docker/Dockerfile.sshagent
```diff
@@ -5,6 +5,9 @@ RUN apk update && \
     apk upgrade && \
     apk add openssh su-exec && \
     rm /var/cache/apk/* && \
+    if [ "$testuid" -gt 0 ]; then if egrep "^[^:]+:x:$testuid:" /etc/passwd; then deluser "$(egrep "^[^:]+:x:$testuid:" /etc/passwd | cut -d: -f1)"; fi; fi && \
+    if [ "$testgid" -gt 0 ]; then if egrep "^[^:]+:x:[^:]+:$testgid:" /etc/passwd; then deluser "$(egrep "^[^:]+:x:[^:]+:$testgid:" /etc/passwd | cut -d: -f1)"; fi; fi && \
+    if [ "$testgid" -gt 0 ]; then if egrep "^[^:]+:x:$testgid:" /etc/group; then delgroup "$(egrep "^[^:]+:x:$testgid:" /etc/group | cut -d: -f1)"; fi; fi && \
```
norrisjeremy, line 10: The issue with the latest Ubuntu 22.04 based Github runners turned out to be that the user inside the Azure VM had a GID of `123`, which conflicted with the built-in `ntp` group inside the Alpine container.
This caused the `addgroup` command below to fail, since it couldn't add a `testuser` group with the same GID.
The gross hack above is to check for any UID/GID conflicts inside the Alpine container and remove the conflicting users and/or groups.

pom.xml
```diff
@@ -92,16 +92,11 @@
...
-            <groupId>com.kohlschutter.junixsocket</groupId>
-            <artifactId>junixsocket-native-common</artifactId>
-            <version>${junixsocket.version}</version>
-            <optional>true</optional>
-        </dependency>
         <dependency>
             <groupId>com.kohlschutter</groupId>
             <artifactId>compiler-annotations</artifactId>
             <version>1.5.2</version>
+            <scope>provided</scope>
```
norrisjeremy, line 99: Using `provided` scope combined with the `includeProvidedScope = false` setting in cyclonedx-maven-plugin prevents a dependency entry from being added to the CycloneDX SBOMs for this, since we only include this to avoid compiler warnings caused by the annotations inside junixsocket.

pom.xml
```diff
@@ -273,7 +277,6 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.10.1</version>
                 <configuration>
-                    <fork>true</fork>
```
norrisjeremy, line 276: For whatever reason, maven-compiler-plugin sometimes doesn't display compiler warnings if it tries to fork a `javac` process.
So I removed it here and opted to just place it inside the `errorprone` profile, since we only need to fork when using errorprone.

Issue #308, opened 2023-04-14
File is getting corrupted when Chunk Mode file transfer interrupted


anamikagsingh:
If there is some network issue while writing the chunk to target directory, the complete chunk will be transferred again and as a result file is getting corrupted.

To solve this, chunk transfer should be resumed instead of transferring the complete chunk again. We added a method checkChunkFailed() in ChannelSftp class and made some changes in our SFTP application which resolved the issue. Here, I have attached ChannelSftp.java file by adding the required changes. Could you please check if the changes can be added to the standard mwiede Jsch library? [ChannelSftp_withChunkModeFix.txt](https://github.com/mwiede/jsch/files/11228924/ChannelSftp_withChunkModeFix.txt)

Please also let me know if it is ok to paste the code directly in the description or is there any other way to do that.
Thanks!

norrisjeremy:
Hi @anamikagsingh,

If you would like us to consider incorporation of your changes, please create a pull request with the proposed changes for us to review.

Thanks,
Jeremy

anamikagsingh:
Hi Jeremy,

I am facing 403 while pushing the changes to this repo, below is the error:
-> "remote: Permission to mwiede/jsch.git denied to anamikagsingh.
fatal: unable to access 'https://github.com/mwiede/jsch.git/': The requested URL returned error: 403"

Please help me with the necessary authorizations.

Regards,
Anamika

mwiede:
you cannot push directly to this repo. Instead you need to create a pull request. see https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request

Issue #302, opened 2023-04-04
UnsupportedOperationException occurs when password prompt is in a different format


anamikagsingh:
java.lang.UnsupportedOperationException occurs when the SFTP server asks for a password prompt in a different format. In one instance of a SFTP Server the prompt was "Enter password:" instead of the traditional "password:".

The issue can be fixed by making changes in file com.jcraft.jsch.UserAuthKeyboardInteractive.java by adding the condition prompt[0].toLowerCase().indexOf("password") >= 0 with OR condition:

The following line 	prompt[0].toLowerCase().indexOf("password:") >= 0){		should be replaced with the below lines:

(prompt[0].toLowerCase().indexOf("password:") >= 0
            		 || prompt[0].toLowerCase().indexOf("password") >= 0)){

mwiede:
the general solution is to implement your own `com.jcraft.jsch.UserInfo` and `com.jcraft.jsch.UIKeyboardInteractive`.
For example this is need, when the server is doing MFA. compare to https://stackoverflow.com/questions/39396209/two-step-verification-using-jsch-library/39511487#39511487

PR #315, opened 2023-04-26
File is getting corrupted when Chunk Mode file transfer interrupted

anamikagsingh:
Hi,

This change will fix the Chunk mode issue raised in https://github.com/mwiede/jsch/issues/308

Regards,
Anamika

### DIFF
--- a/src/main/java/com/jcraft/jsch/ChannelSftp.java
+++ b/src/main/java/com/jcraft/jsch/ChannelSftp.java
@@ -648,7 +648,7 @@ public class ChannelSftp extends ChannelSession {
       throw new SftpException(SSH_FX_FAILURE, e.toString(), e);
     }
   
-	public boolean checkChunkFailed(InputStream src, String dst, Long startChunk)throws SftpException{
+	public boolean checkChunkFailed(String dst, Long startChunk)throws SftpException{
 	  try{
 	      ((MyPipedInputStream)io_in).updateReadSide();
 
--- a/src/main/java/com/jcraft/jsch/ChannelSftp.java
+++ b/src/main/java/com/jcraft/jsch/ChannelSftp.java
@@ -32,6 +32,9 @@ import java.nio.charset.StandardCharsets;
 import java.util.Hashtable;
 import java.util.Vector;
 
+import com.jcraft.jsch.Channel.MyPipedInputStream;
+
+
 public class ChannelSftp extends ChannelSession {
 
   static private final int LOCAL_MAXIMUM_PACKET_SIZE = 32 * 1024;
@@ -158,6 +161,7 @@ public class ChannelSftp extends ChannelSession {
   private boolean fEncoding_is_utf8 = true;
 
   private RequestQueue rq = new RequestQueue(16);
+  private String chunkStart = null;
 
   /**
    * Specify how many requests may be sent at any one time. Increasing this value may slightly
@@ -516,7 +520,13 @@ public class ChannelSftp extends ChannelSession {
           // System.err.println(eee);
         }
       }
-      if (mode == RESUME && skip > 0) {
+    if((mode==RESUME&& chunkStart!=null) && skip>0){
+  	    long skipped=src.skip(skip-Long.parseLong(chunkStart));
+  	    if(skipped<(skip-Long.parseLong(chunkStart))){
+  	    	throw new SftpException(SSH_FX_FAILURE, "failed to resume for "+dst);
+  	    }
+  	 }
+  	else if(mode==RESUME && skip>0){
         long skipped = src.skip(skip);
         if (skipped < skip) {
           throw new SftpException(SSH_FX_FAILURE, "failed to resume for " + dst);
@@ -643,6 +653,32 @@ public class ChannelSftp extends ChannelSession {
         throw (SftpException) e;
       throw new SftpException(SSH_FX_FAILURE, e.toString(), e);
     }
+  
+  public boolean checkChunkFailed(InputStream src, String dst, String ChunkStart)throws SftpException{
+	  try{
+	      ((MyPipedInputStream)io_in).updateReadSide();
+
+	      byte[] dstb=Util.str2byte(dst, fEncoding);
+	      long skip=0;
+	      try{
+		  SftpATTRS attr=_stat(dstb);
+		  skip=attr.getSize();
+	      }
+	      catch(Exception eee){
+		  //System.err.println(eee);
+	      }
+	      
+	      if(skip>0){
+	    	  if(skip == Long.parseLong(ChunkStart))
+	    		  return false;
+	    	  else return true;
+	      }
+	      else return false;
+	  }
+	  catch(Exception e){
+		  throw new SftpException(SSH_FX_FAILURE, "Cannot check-" + e.toString());
+	  }
+	  
   }
 
   public OutputStream put(String dst) throws SftpException {
@@ -2952,6 +2988,10 @@ public class ChannelSftp extends ChannelSession {
         throw (SftpException) e;
       throw new SftpException(SSH_FX_FAILURE, e.toString(), e);
     }
+  }	  
+
+  public void setChunkStart(String chunkStart) {
+	  this.chunkStart= chunkStart;
   }
 
   public static class LsEntry implements Comparable<LsEntry> {
--- a/src/main/java/com/jcraft/jsch/ChannelSftp.java
+++ b/src/main/java/com/jcraft/jsch/ChannelSftp.java
@@ -32,7 +32,6 @@ import java.nio.charset.StandardCharsets;
 import java.util.Hashtable;
 import java.util.Vector;
 
-import com.jcraft.jsch.Channel.MyPipedInputStream;
 
 
 public class ChannelSftp extends ChannelSession {
@@ -161,7 +160,7 @@ public class ChannelSftp extends ChannelSession {
   private boolean fEncoding_is_utf8 = true;
 
   private RequestQueue rq = new RequestQueue(16);
-  private String chunkStart = null;
+  private Long chunkStart = null;
 
   /**
    * Specify how many requests may be sent at any one time. Increasing this value may slightly
@@ -520,15 +519,10 @@ public class ChannelSftp extends ChannelSession {
           // System.err.println(eee);
         }
       }
-    if((mode==RESUME&& chunkStart!=null) && skip>0){
-  	    long skipped=src.skip(skip-Long.parseLong(chunkStart));
-  	    if(skipped<(skip-Long.parseLong(chunkStart))){
-  	    	throw new SftpException(SSH_FX_FAILURE, "failed to resume for "+dst);
-  	    }
-  	 }
-  	else if(mode==RESUME && skip>0){
-        long skipped = src.skip(skip);
-        if (skipped < skip) {
+    if (mode == RESUME && skip > 0) {
+          long srcSkip = chunkStart != null ? skip - chunkStart : skip;
+          long skipped = src.skip(srcSkip);
+          if (skipped < srcSkip) {
           throw new SftpException(SSH_FX_FAILURE, "failed to resume for " + dst);
         }
       }
@@ -654,7 +648,7 @@ public class ChannelSftp extends ChannelSession {
       throw new SftpException(SSH_FX_FAILURE, e.toString(), e);
     }
   
-  public boolean checkChunkFailed(InputStream src, String dst, String ChunkStart)throws SftpException{
+	public boolean checkChunkFailed(InputStream src, String dst, Long startChunk)throws SftpException{
 	  try{
 	      ((MyPipedInputStream)io_in).updateReadSide();
 
@@ -669,7 +663,7 @@ public class ChannelSftp extends ChannelSession {
 	      }
 	      
 	      if(skip>0){
-	    	  if(skip == Long.parseLong(ChunkStart))
+	    	  if(skip == startChunk)
 	    		  return false;
 	    	  else return true;
 	      }
--- a/src/main/java/com/jcraft/jsch/Session.java
+++ b/src/main/java/com/jcraft/jsch/Session.java
@@ -129,6 +129,8 @@ public class Session {
 
   Buffer buf;
   Packet packet;
+  
+  List<String> fingerprintList = new ArrayList<String>();
 
   SocketFactory socket_factory = null;
 
@@ -883,9 +885,20 @@ public class Session {
       }
       // System.err.println("finger-print: "+key_fprint);
       if (userinfo != null) {
+		if (shkc.equals("ask")) {
         boolean foo = userinfo.promptYesNo("The authenticity of host '" + chost
             + "' can't be established.\n" + key_type + " key fingerprint is " + key_fprint + ".\n"
             + "Are you sure you want to continue connecting?");
+		} else {
+		boolean foo = false; 
+    	  for(String eachFingerPrint : fingerprintList){
+    		  if(key_fprint.equalsIgnoreCase(eachFingerPrint))
+    		  {
+    			  foo = true;
+    			  break;
+    		  }
+    	  }	
+		}
         if (!foo) {
           throw new JSchException("reject HostKey: " + chost);
         }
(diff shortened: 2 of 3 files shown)


norrisjeremy:
Hi @anamikagsingh,

1. Can you execute `mvn formatter:format` so that the code changes are formatted correctly?
2. Can you provide more details on the issue you are seeing and how your proposed changes help to resolve it?
-- Specifically, can you describe the overall interaction that a JSch consumer would utilize in order to make use of the new `checkChunkFailed()` & `setChunkStart()` methods?
-- It may help if you can provide some short code examples that depict how the file corruption can occur during an interrupted transfer, and then how a user would mitigate the file corruption using the proposed changes.

Thanks,
Jeremy

anamikagsingh:
Hi Jeremy,

I also have added changed files 'Session.java' and 'UserAuthKeyboardInteractive.java' to this pull request. This change is to fix the issue raised in #302 and also to support multiple server fingerprint.

Regards,
Anamika

norrisjeremy:
Hi @anamikagsingh,

1. You need to execute `mvn formatter:format` and commit, then push the resulting changes to this PR.
2. Can you please split the changes to `Session.java` and `UserAuthKeyboardInteractive.java` into individual commits, instead of a single combined commit?
3. Can you modify your commit messages to be more descriptive than simply `ChannelSftp` and `SessionAndUserAuthKeyboardInteractive`?
4. I'm not sure I follow what you mean by `we have an option to transfer large files in smaller chunks and append them in the target in the sequence they have been transferred`: how are you transferring these large files via smaller chunks using JSch's existing `ChannelSftp` API methods?
-- I also do not understand how an application would determine the `chunkStart` value to use for the `checkChunkFailed()` method: can you better explain how you are determining this value?
5. I'm not sure I understand the purpose of the changes to `Session.java`: can't you implement this custom logic of authenticating hosts by their host key fingerprint by simply implementing your own custom `HostKeyRepository` and then instructing JSch use it (via either `jsch.setHostKeyRepository()` or `session.setHostKeyRepository()`)?
6. I'm not sure I understand the proposed changes to `UserAuthKeyboardInteractive.java`: why not simply implement a custom `UserInfo` as suggested in #302?
-- Also, you mentioned in #302 that you had a server that presented a prompt of `Enter password:`, but doesn't that prompt already work without any changes since that should already match the existing `prompt[0].toLowerCase().indexOf("password:") >= 0`?

Thanks,
Jeremy

src/main/java/com/jcraft/jsch/Session.java
```diff
@@ -883,9 +885,20 @@ private void checkHost(String chost, int port, KeyExchange kex) throws JSchExcep
       }
       // System.err.println("finger-print: "+key_fprint);
       if (userinfo != null) {
+		if (shkc.equals("ask")) {
         boolean foo = userinfo.promptYesNo("The authenticity of host '" + chost
             + "' can't be established.\n" + key_type + " key fingerprint is " + key_fprint + ".\n"
             + "Are you sure you want to continue connecting?");
+		} else {
```
norrisjeremy, line 892: This appears to be a dead branch because up above we already throw a JSchException if `shkc.equals("yes")`.

src/main/java/com/jcraft/jsch/Session.java
```diff
@@ -3198,6 +3211,12 @@ private void checkConfig(ConfigRepository.Config config, String key) {
     if (value != null)
       this.setConfig(key, value);
   }
+  
+  public void addToFingerprintList(String channelFingerprint)
+	{
+		String []fingerprints = channelFingerprint.split("~");
```
norrisjeremy, line 3217: Is there some standard you can point to that has SSH host key fingerprints concatenated together via `~` characters?
Or is this some sort of locally invented format?

src/main/java/com/jcraft/jsch/UserAuthKeyboardInteractive.java
```diff
@@ -129,7 +129,8 @@ public boolean start(Session session) throws Exception {
           byte[][] response = null;
 
           if (password != null && prompt.length == 1 && !echo[0]
-              && prompt[0].toLowerCase().indexOf("password:") >= 0) {
+              && (prompt[0].toLowerCase().indexOf("password:") >= 0
+            		 || prompt[0].toLowerCase().indexOf("password") >= 0)){
```
norrisjeremy, line 133: In #302 you mention that this change is needed in order to support a server that has a prompt of `Enter password:`.
However, the string `Enter password:` already should match `prompt[0].toLowerCase().indexOf("password:") >= 0`, so I'm not sure I understand the purpose of this change (see below)?

```
jshell> String foo = "Enter password:";
foo ==> "Enter password:"

jshell> if (foo.toLowerCase().indexOf("password:") >= 0) System.out.println("matches"); else throw new Exception("fail");
matches

jshell> 
```

src/main/java/com/jcraft/jsch/ChannelSftp.java
```diff
@@ -643,6 +653,32 @@ public void _put(InputStream src, String dst, SftpProgressMonitor monitor, int m
         throw (SftpException) e;
       throw new SftpException(SSH_FX_FAILURE, e.toString(), e);
     }
+  
+  public boolean checkChunkFailed(InputStream src, String dst, String ChunkStart)throws SftpException{
```
norrisjeremy, line 657: 1. The argument `InputStream src` appears to be unused in this method, so why is it needed?
2. The argument `String ChunkStart` should be changed to not start with a capital letter.
-- Perhaps something like `String chunkStart` instead?

src/main/java/com/jcraft/jsch/ChannelSftp.java
```diff
@@ -643,6 +653,32 @@ public void _put(InputStream src, String dst, SftpProgressMonitor monitor, int m
...
+	      
+	      if(skip>0){
+	    	  if(skip == Long.parseLong(ChunkStart))
+	    		  return false;
+	    	  else return true;
+	      }
+	      else return false;
+	  }
+	  catch(Exception e){
+		  throw new SftpException(SSH_FX_FAILURE, "Cannot check-" + e.toString());
```
norrisjeremy, line 679: 1. We should make sure `e` isn't already an instance of `SftpException` first, and if it is, just rethrow it directly.
2. We should also add the original exception as the cause of the new `SftpException`.

```
if (e instanceof SftpException)
  throw (SftpException) e;
throw new SftpException(SSH_FX_FAILURE, e.toString(), e);
```

src/main/java/com/jcraft/jsch/ChannelSftp.java
```diff
@@ -32,6 +32,9 @@
 import java.util.Hashtable;
 import java.util.Vector;
 
+import com.jcraft.jsch.Channel.MyPipedInputStream;
```
norrisjeremy, line 35: This is unnecessary.
Since `ChannelSftp` extends `Channel`, it can directly reference the `MyPipedInputStream` class without an import statement.

src/main/java/com/jcraft/jsch/ChannelSftp.java
```diff
@@ -516,7 +520,13 @@ public void _put(InputStream src, String dst, SftpProgressMonitor monitor, int m
...
         }
       }
-      if (mode == RESUME && skip > 0) {
+    if((mode==RESUME&& chunkStart!=null) && skip>0){
+  	    long skipped=src.skip(skip-Long.parseLong(chunkStart));
+  	    if(skipped<(skip-Long.parseLong(chunkStart))){
+  	    	throw new SftpException(SSH_FX_FAILURE, "failed to resume for "+dst);
+  	    }
+  	 }
+  	else if(mode==RESUME && skip>0){
```
norrisjeremy, line 529: I think this could be better written as simply something like this to avoid code duplication:
```
      if (mode == RESUME && skip > 0) {
        long srcSkip = chunkStart != null ? skip - Long.parseLong(chunkStart) : skip;
        long skipped = src.skip(srcSkip);
        if (skipped < srcSkip) {
          throw new SftpException(SSH_FX_FAILURE, "failed to resume for " + dst);
        }
      }
```

src/main/java/com/jcraft/jsch/ChannelSftp.java
```diff
@@ -158,6 +161,7 @@ public class ChannelSftp extends ChannelSession {
   private boolean fEncoding_is_utf8 = true;
 
   private RequestQueue rq = new RequestQueue(16);
+  private String chunkStart = null;
```
norrisjeremy, line 164: Why does `chunkStart` need to be a `String` instead of having it simply be a `long`?
This would avoid a lot of unwieldy calls to `Long.parseLong(chunkStart)`?

src/main/java/com/jcraft/jsch/Session.java
```diff
@@ -883,9 +885,20 @@ private void checkHost(String chost, int port, KeyExchange kex) throws JSchExcep
...
+		} else {
+		boolean foo = false; 
+    	  for(String eachFingerPrint : fingerprintList){
+    		  if(key_fprint.equalsIgnoreCase(eachFingerPrint))
+    		  {
+    			  foo = true;
+    			  break;
+    		  }
+    	  }	
+		}
```
norrisjeremy, line 901: Also, can you point at any other applications or libraries that perform SSH host key verification using fingerprints?
This seems like a pretty unusual (and potentially insecure) practice that I am unaware of, and I'm not sure we want to directly enable this sort of thing natively within JSch.

src/main/java/com/jcraft/jsch/Session.java
```diff
@@ -129,6 +129,8 @@ public class Session {
 
   Buffer buf;
   Packet packet;
+  
+  List<String> fingerprintList = new ArrayList<String>();
```
norrisjeremy, line 133: This can be shortened to simply `new ArrayList<>()` instead of `new ArrayList<String>()`.

norrisjeremy (changes requested):

src/main/java/com/jcraft/jsch/Session.java
anamikagsingh: I have changed the separator to " , " now.

src/main/java/com/jcraft/jsch/UserAuthKeyboardInteractive.java
anamikagsingh: I am not able to reproduce this issue in my test system so cannot test it. We can leave the changes done in UserAuthKeyboardInteractive.java for now. Please let me know if I need to revert the changes and commit again. Thanks.

src/main/java/com/jcraft/jsch/ChannelSftp.java
anamikagsingh: 1. Removed the argument InputStream src from the method.
2. Changed the parameter name to startChunk

src/main/java/com/jcraft/jsch/ChannelSftp.java
anamikagsingh: Removed the import as it was not needed. Thanks for the suggestions.

src/main/java/com/jcraft/jsch/ChannelSftp.java
anamikagsingh: I have made the changes as suggested.

src/main/java/com/jcraft/jsch/ChannelSftp.java
anamikagsingh: Changed it to Long now and made the other changes accordingly.

src/main/java/com/jcraft/jsch/Session.java
anamikagsingh: I have no idea if any other application which is using Jsch library is also using fingerprints to perform SSH host key verification. It is being used in SAP PI SFTP Adapter.

src/main/java/com/jcraft/jsch/Session.java
anamikagsingh: Changed it as suggested.

norrisjeremy:
HI @anamikagsingh,

Can you please answer the following questions?

1. I'm not sure I follow what you mean by `we have an option to transfer large files in smaller chunks and append them in the target in the sequence they have been transferred`: how are you transferring these large files via smaller chunks using JSch's existing `ChannelSftp` API methods? Can you provide some example code of how you are using `ChannelSftp` and how your proposed API changes are utilized to remedy the file corruption issue you are having?
2. I'm not sure I understand the purpose of the changes to `Session.java`: is there a reason you cannot simply achieve this custom logic of authenticating hosts by their host key fingerprint by simply implementing your own custom HostKeyRepository and then instructing JSch use it (via either `jsch.setHostKeyRepository()` or `session.setHostKeyRepository()`)?

Thanks,
Jeremy

anamikagsingh:
> HI @anamikagsingh,
> 
> Can you please answer the following questions?
> 
> 1. I'm not sure I follow what you mean by `we have an option to transfer large files in smaller chunks and append them in the target in the sequence they have been transferred`: how are you transferring these large files via smaller chunks using JSch's existing `ChannelSftp` API methods? Can you provide some example code of how you are using `ChannelSftp` and how your proposed API changes are utilized to remedy the file corruption issue you are having?
> 2. I'm not sure I understand the purpose of the changes to `Session.java`: is there a reason you cannot simply achieve this custom logic of authenticating hosts by their host key fingerprint by simply implementing your own custom HostKeyRepository and then instructing JSch use it (via either `jsch.setHostKeyRepository()` or `session.setHostKeyRepository()`)?
> 
> Thanks, Jeremy

About point 1, when chunkmode is enabled, we set the chunk size (i.e. 10 MB). Based on chunk size, the bytes will be read from the input file and processed further. We are using the below snippet for chunk mode transfer in Adapter:
```

  if(chunkMode) {					
	long readBytes = 0;
	String user = propHandler.getPropertyValueAsString(ChannelConstants.AUTH_USERNAME);
	String chunkID = getChunkID(con.getHost(), user, fileAtts);
	//for recovery in chunk mode
	boolean retry = false;
	String chunkEntry = checkChunkStatus(chunkID);
	if(chunkEntry != null) {
		readBytes = Long.parseLong(chunkEntry);
		if(readBytes > 0) {
		retry = true;
		 }
		}				
		//read from inputstream
		byte[] chunk = new byte[chunkSize];
		int curChunkSize = -1;
		curChunkSize = readChunk(chunk, is, chunkSize);
		TRACE.debugT(SIGNATURE, "Bytes read: " + curChunkSize);
		if(curChunkSize >0 || processEmptyFile ==true) {
		audit.addAuditLogEntry(msg.getMessageKey(), AuditLogStatus.SUCCESS, "Starting to read file from byte: " + readBytes);
		readBytes += curChunkSize;
		TRACE.debugT(SIGNATURE, "Total read Bytes: " + readBytes);

		TxTicket ticket = null;
		//sending message to XI
		try {
			//create transaction
			ticket = createTransaction(channelDeliverySemantics);
			updateChunkStatus(chunkID, readBytes + "");
			if(curChunkSize == chunkSize) {
				sendMessageFromSftp(chunk, fileAtts, attachments, channel, con, home, msg,isSpecialFileHandling, specialFileStorageLocation);
			} else {
				byte[] trimChunk = new byte[curChunkSize];
				System.arraycopy(chunk, 0, trimChunk, 0, curChunkSize);
				sendMessageFromSftp(trimChunk, fileAtts, attachments, channel, con, home, msg,isSpecialFileHandling, specialFileStorageLocation);
				
				//final chunk; hence add it to message ID mapper for duplicate check
				addToIDMapper(con.getHost(), fileAtts, msg.getMessageKey().getMessageId());
			}
			if(ticket != null) {
				commitTransaction(ticket);
			}
			xiMsgSent = true;
		} catch(MessagingException e) {
		 }
	}                               
```
About point 2, I will try to figure it out.

Thanks.

norrisjeremy:
Hi @anamikagsingh,

From what I can understand, this `chunkmode` handling seems to be specific to your application.
As such, I'm not sure if I believe it is appropriate to merge it into JSch, lest it introduce regressions for other JSch users.

Thanks,
Jeremy

src/main/java/com/jcraft/jsch/UserAuthKeyboardInteractive.java
ecki: The or also overlaps, you could just remove the colon..

Issue #330, opened 2023-05-24
Improve error handling in InputStream.close() for SFTP channels


stsiano:
For most _sendClose() calls in ChannelSFftp the errors are handled just fine. However there is one exception:
When an exception is happening in the close() method of the anonymous InputStream() implementation in 
https://github.com/mwiede/jsch/blob/master/src/main/java/com/jcraft/jsch/ChannelSftp.java#L1497
the Exception that was causing the error will be swallowed and replaced by an IOException with the error text "error".

This makes error analysis really difficult in these cases. Is it possible to preserve the original exception somehow (as in all the other calls of _sendClose()?

PR #331, opened 2023-05-24
Improve error handling in InputStream.close() for SFTP channels

stsiano:
Fixes https://github.com/mwiede/jsch/issues/330

### DIFF
--- a/src/main/java/com/jcraft/jsch/ChannelSftp.java
+++ b/src/main/java/com/jcraft/jsch/ChannelSftp.java
@@ -1495,8 +1495,10 @@ public class ChannelSftp extends ChannelSession {
           rq.cancel(header, buf);
           try {
             _sendCLOSE(handle, header);
+          } catch (IOException e) {
+            throw e;
           } catch (Exception e) {
-            throw new IOException("error");
+            throw new IOException(e.toString(), e);
           }
         }
       };


src/main/java/com/jcraft/jsch/ChannelSftp.java
```diff
@@ -1495,8 +1495,10 @@ public void close() throws IOException {
           rq.cancel(header, buf);
           try {
             _sendCLOSE(handle, header);
+          } catch (IOException e) {
+            throw e;
           } catch (Exception e) {
-            throw new IOException("error");
+            throw new IOException(e.toString(), e);
```
norrisjeremy, line 1501: I would suggest doing it like this instead:
```
if (e instanceof IOException)
  throw (IOException) e;
throw new IOException(e.toString(), e);
```

norrisjeremy (changes requested):

src/main/java/com/jcraft/jsch/ChannelSftp.java
stsiano: I copied this particular code block from another part of the same file. 

Sonar proposes to do this the way I did this
https://rules.sonarsource.com/java/RSPEC-1193

In the end the code is equivalent, so if you care about it, I  can also change the pull request.

norrisjeremy: Ok, I see that now, apologies.

norrisjeremy (approved):

PR #344, opened 2023-06-09
Make OSGi dependencies optional

OndroMih:
This will make it possible to use the jsch artifact as an OSGi module without another OSGi module(s) providing optional dependencies. All those dependencies are already marked as optional in the JPMS module-info.java file: https://github.com/mwiede/jsch/blob/master/src/main/java9/module-info.java. They should be also optional in the OSGi environment.

This PR will also allow using the Jsch as a dependency in Eclipse GlassFish, without repackaging it into a different OSGi package. With this merged into the upstream project, the following PR won't be necessary:  https://github.com/eclipse-ee4j/glassfish-repackaged/pull/21

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -403,10 +403,10 @@
                     <instructions>
                         <Export-Package>com.jcraft.jsch;-noimport:=true</Export-Package>
                         <Import-Package><![CDATA[
-                            com.sun.jna*;version="${range;[=0,+)}",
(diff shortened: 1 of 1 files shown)


OndroMih:
Yes, you are right, @norrisjeremy. Thanks for your fast reply, I'm sorry to bother you. I think I was using an older version of jsch before, which set the OSGi dependencies as required.

Thanks for your support and I wish you a lovely weekend!

Issue #392, opened 2023-09-14
No virtual method intValueExact() in Android API 30 and below


deskangel:
```java
  static int parseASN1IntegerAsInt(byte[] content) {
    return new BigInteger(content).intValueExact();
  }
```

It looks like jsch may not be compatible with older Android versions (xxxExact was introduced in Android API 31) based on the code above. Are there any recommended workarounds for this?

norrisjeremy:
Hi @deskangel,

I don't have any recommended workarounds other than to use a Java implementation that fully implements the Java 8 specification, since this JSch fork requires Java 8+ to function.

You are also welcome to submit a PR that we can review to attempt to remedy this issue if you believe you can find a way to implement the above method in a different way that is compatible with the non fully compliant version of Java you are attempting to use.

Thanks,
Jeremy

deskangel:
Hi @norrisjeremy,

I don't have the development environment for jsch and Java, so it's not very convenient for me to submit a PR. The reason I raised this issue is that I don't think it's necessary to introduce this function; It simply throws an exception when the value exceeds the int range compared to intValue. Previous versions of jsch worked well on older versions of Android, so I feel a bit regretful that it is incompatible with older Android versions just because of this function. 

That's all.

mwiede:
@deskangel I can recommend Github Codespaces, which gives fast access to any repository out there and you could start working directly in the browser...

I have looked at this, but even [desugaring ](https://developer.android.com/studio/write/java8-support) did not help.
After compiling jsch in Android studio, it seems that this line of code is the only incompatibility with the Android API, so I have will create a small PR.

PR #397, opened 2023-09-26
#392 replace call to BigInteger.intValueExact to remain comptaible wi…

mwiede:
Although java 8 compatibility is given, it is just a one line and an easy fix.


norrisjeremy:
I really dislike having to implement stuff like this because Google produces non-standard Java implementations for Android.

norrisjeremy (approved):
Reluctant approval...

PR #409, opened 2023-10-21
Reset the sequence numbers on Session disconnect to support reconnection

### LINKED ISSUE #408: Reconnecting after JSchAlgoNegoFailException with new server_host_key results in corrupt packets
Using `0.1.72` we expect to receive an `Auth fail` exception when connecting to a legacy system. While upgrading to `0.2.11` we implemented a naive retry using the README :
```java
        JSch jsch = new JSch();
        Session session = jsch.getSession(user, host, port);
        try {
            session.connect();
        } catch (JSchException e) {
            session.setConfig("server_host_key", session.getConfig("server_host_key") + ",ssh-rsa");
            session.setConfig("PubkeyAcceptedAlgorithms", session.getConfig("PubkeyAcceptedAlgorithms") + ",ssh-rsa");

            // FIXME : Reconnect using deprecated algo , accept new keys or ...
            session.connect();
        }
```

```
Algorithm negotiation fail: algorithmName=\"server_host_key\"

SSH_MSG_NEWKEYS sent
SSH_MSG_NEWKEYS received
SSH_MSG_SERVICE_REQUEST sent
Caused by: com.jcraft.jsch.JSchException: Packet corrupt
	at com.jcraft.jsch.Session.start_discard(Session.java:1367) ~[jsch-0.2.0.jar:0.2.0]
	at com.jcraft.jsch.Session.read(Session.java:1223) ~[jsch-0.2.0.jar:0.2.0]
	at com.jcraft.jsch.UserAuthNone.start(UserAuthNone.java:57) ~[jsch-0.2.0.jar:0.2.0]
	at com.jcraft.jsch.Session.connect(Session.java:401) ~[jsch-0.2.0.jar:0.2.0]
	at com.jcraft.jsch.Session.connect(Session.java:185) ~[jsch-0.2.0.jar:0.2.0]
```

When we configure `ssh-rsa` to be used in the initial connection all works fine , only on the retry it fails with packet corruption and have seen a few `End of IO Stream Read` but not entirely sure if that was related. This indicates to me that there is some state preserved between the initial `connect()` which on failure does a best effort `disconnect()` and then with the new config a `connect()` attempt.

mvegter:
Resolves #408

### DIFF
--- a/src/test/java/com/jcraft/jsch/SessionReconnectIT.java
+++ b/src/test/java/com/jcraft/jsch/SessionReconnectIT.java
@@ -1,0 +1,160 @@
+package com.jcraft.jsch;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+import com.github.valfirst.slf4jtest.LoggingEvent;
+import com.github.valfirst.slf4jtest.TestLogger;
+import com.github.valfirst.slf4jtest.TestLoggerFactory;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Base64;
+import java.util.List;
+import java.util.Random;
+import org.apache.commons.codec.digest.DigestUtils;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.output.Slf4jLogConsumer;
+import org.testcontainers.images.builder.ImageFromDockerfile;
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+@Testcontainers
+public class SessionReconnectIT {
+
+  private static final int timeout = 2000;
+  private static final DigestUtils sha256sum = new DigestUtils(DigestUtils.getSha256Digest());
+  private static final TestLogger jschLogger = TestLoggerFactory.getTestLogger(JSch.class);
+  private static final TestLogger sshdLogger =
+      TestLoggerFactory.getTestLogger(SessionReconnectIT.class);
+
+  @TempDir
+  public Path tmpDir;
+  private Path in;
+  private Path out;
+  private String hash;
+  private Slf4jLogConsumer sshdLogConsumer;
+
+  @Container
+  public GenericContainer<?> sshd = new GenericContainer<>(new ImageFromDockerfile()
+      .withFileFromClasspath("dropbear_rsa_host_key", "docker/dropbear_rsa_host_key")
+      .withFileFromClasspath("authorized_keys", "docker/authorized_keys")
+      .withFileFromClasspath("Dockerfile", "docker/Dockerfile.dropbear")).withExposedPorts(22);
+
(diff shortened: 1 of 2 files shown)


mvegter:
Can also be reproduced by changing any of the Algorithms IT to do :
```java
    try {
      session.connect(timeout);
      assertTrue(session.isConnected());

      // Verify reconnect
      session.disconnect();
      session.connect(timeout);
      assertTrue(session.isConnected());

    } finally {
      session.disconnect();
    }
```

src/test/java/com/jcraft/jsch/SessionReconnectIT.java
```diff
@@ -0,0 +1,161 @@
...
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Base64;
+import java.util.List;
+import java.util.Random;
```
norrisjeremy, line 25: Can you organize the imports so that they abide by [Google Java Style Guide #3.3.3](https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing)?

norrisjeremy (changes requested):

src/test/java/com/jcraft/jsch/SessionReconnectIT.java
```diff
@@ -0,0 +1,160 @@
...
+    try {
+      doSftp(session, false);
+      fail("exception expected");
+    } catch (JSchAlgoNegoFailException e) {
+      // Dropbear does not support rsa-sha2-512/rsa-sha2-256, so add ssh-rsa
+      String serverHostKey = session.getConfig("server_host_key") + ",ssh-rsa";
+      String pubkeyAcceptedAlgorithms = session.getConfig("PubkeyAcceptedAlgorithms") + ",ssh-rsa";
+      session.setConfig("server_host_key", serverHostKey);
+      session.setConfig("PubkeyAcceptedAlgorithms", pubkeyAcceptedAlgorithms);
+      doSftp(session, true);
```
norrisjeremy, line 100: I wonder if we should wrap this in a `assertDoesNotThrow()` to emphasize that we expect this to succeed?

mvegter: The test already fails if an Exception is thrown, so right now it is indeed implied. Let me know if you want that changed or not , i don't really see that as a common pattern in the current ITs

norrisjeremy (commented):
I wouldn't be surprised if the intention was for the Session object to not be reusable to create a new connection after the disconnect method has been called. Why not instead simply create a new Session object instead of trying to reuse the existing one?

Also considering this isn't a regression per se (since JSch has behaved in this manner for years), I would lean more towards not merging this in, since there is always a chance that this could instead introduce a some other new regression of some sort instead.

@mwiede what do you think?

mvegter:
> I wouldn't be surprised if the intention was for the Session object to not be reusable to create a new connection after the disconnect method has been called. Why not instead simply create a new Session object instead of trying to reuse the existing one?

I understand your position on this change, however the reason for introducing this is that the Session already resets the `isConnected` flag. This technically allows anyone to call `connect()` again , whereas a connected session throws an Exception if we attempt to re-connect. If we were to disallow this I would expect (perhaps in a future change) to throw an Exception when calling `connect()` after a `disconnect()`

On another note, allowing a client to call `connect()` multiple times can also allow for more resilience with retries on certain IOException without setting up a new Jsch/Session .

norrisjeremy (approved):
I don't feel strongly enough about this to block it, as long as @mwiede is also ok with it.

mwiede:
we do not really know, whether the Session was intened for re-use or not, but I think we can accept this change, because it also cannot be seen in the code, that reuse has to be blocked.

Issue #357, opened 2023-07-09
Throw specific exceptions at least for connect and authentication


erik-wramner:
JSch always throws the same checked exception for any errors. In some cases that is fine, but in others it is a pain. Specifically, when connecting to a host, it would be nice to know if the connection fails because the host does not exist or is down or if a TCP connection was established and something went wrong in a later stage. It would also be very useful to know if the connection failed due to an invalid user or password, as the application should typically not retry (again - the library itself retries authentications) as that can lock the account.

Gradually converting the code to throw sub-classes of the existing exception with known meanings would be a great improvement.

PR #416, opened 2023-10-24
Introduce JSchSessionDisconnectException to allow the reasonCode to be retrieved without String parsing

mvegter:
Similar to #410 , did not change any of the messages in case anyone is matching the exception message due to the lack of specific sub exception throwing .

Relates to #357

### DIFF
--- a/src/main/java/com/jcraft/jsch/JSchSessionDisconnectException.java
+++ b/src/main/java/com/jcraft/jsch/JSchSessionDisconnectException.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright (c) 2002-2018 ymnk, JCraft,Inc. All rights reserved.
+ *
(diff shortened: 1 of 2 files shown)


src/main/java/com/jcraft/jsch/JSchSessionDisconnectException.java
```diff
@@ -0,0 +1,49 @@
...
+package com.jcraft.jsch;
+
+public class JSchSessionDisconnectException extends JSchException {
+  private static final long serialVersionUID = -1L;
+
+  // RFC 4250 4.2.2.
+  private final int reasonCode;
+  private final String description;
+
+  JSchSessionDisconnectException(String s, int reasonCode, String description) {
```
norrisjeremy, line 36: Any reason to not also include the language tag?

norrisjeremy (approved):

Issue #411, opened 2023-10-22
Upload large file getting stuck


mironabr:
I was using the original jsch lib and had the issue that uploading large file getting stack and I applied the following fix to the code and it solved the issue.
I want to move to your lib but you must apply this fix as well:
https://github.com/is/jsch/pull/39

norrisjeremy:
Hi @mironabr,

Please see #54 that the original author of https://github.com/is/jsch/pull/39 opened here.
The fix proposed didn't seem to fully resolve the issue based upon my testing at the time.
I'm also extremely hesitant to merge the proposed fix as I'm not convinced it wouldn't introduce regressions for some folks.

Thanks,
Jeremy

mironabr:
We moved from the original jsch to this lib. It works. However, our QA team made performance tests and said that uploading file to SFTP server with the original lib is much faster. Do you know something about it ?

mironabr:
My QA team did lot of tests with the original jsch lib and mwiede lib and they 100% confirmed that uploading large file (20MB and above), most of the times both libs are getting stuck but when I added the above fix (https://github.com/is/jsch/pull/39) both libs worked and were not stuck.
Also the QA team confirmed that uploading using the original jsch lib is much faster

norrisjeremy:
Hi @mironabr,

I don't know immediately why your team would have found that this fork is slower than the older version of than this version, since you haven't provided any data that can be analyzed to help determine why.
If I had to guess, my initial suspicion would because our version may using different crypto algorithms than the original with your server, since we have added several more modern and secure crypto algorithms into this fork, that likely have different performance characteristics.

As for the upload issue, I'm not especially comfortable with merging the fix, since I was still able to trigger a deadlock even with it applied, and the original author of it has remained silent and never answered questions I posed to help better understand the proposed fix.

Thanks,
Jeremy

norrisjeremy:
@mwiede What are your thoughts? Do you think we should simply merge the change from https://github.com/is/jsch/pull/39 here? I'm not sure how much of a regression risk there would be for it, but if users are reporting it helps solve their problems, perhaps it's worth a gamble.

norrisjeremy:
> @mwiede What are your thoughts? Do you think we should simply merge the change from [is/jsch#39](https://github.com/is/jsch/pull/39) here? I'm not sure how much of a regression risk there would be for it, but if users are reporting it helps solve their problems, perhaps it's worth a gamble.

[...]

PR #425, opened 2023-11-05
Update ChannelSftp.java

mironabr:
flush

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2,7 +2,7 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <groupId>com.github.mwiede</groupId>
-    <artifactId>jsch</artifactId>
+    <artifactId>k2-jsch</artifactId>
     <packaging>jar</packaging>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -2,7 +2,7 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <groupId>com.github.mwiede</groupId>
-    <artifactId>jsch</artifactId>
+    <artifactId>k2-jsch</artifactId>
```
mwiede, line 5: I think this was added by mistake right?

mwiede:
@mironabr as said in #411 can we have a failing integration test which this PR would solve?

pom.xml
mironabr: Right. Currently we are using our fork in our code.

mironabr:
> @mironabr as said in #411 can we have a failing integration test which this PR would solve?

It is difficult to do that.  The issue does not happen any time. I does not happen on any SFTP server. I guess it depend on the SFTP server type/configuration. It happen on our local test servers. I saw lot of talking about it over the internet

mwiede:
we took the change in [86bc2e6](https://github.com/mwiede/jsch/pull/431/commits/86bc2e6ef86d9bc8137887717149e7fa80d2cb22) so I will close this PR

Issue #403, opened 2023-10-05
channel.getInputStream() should warn if the channel has already connected


pshem:
We recently fixed a bug in our code where we were doing two things in the wrong order, which would lead to a race condition.

Examples, such as https://github.com/mwiede/jsch/blob/c6970b8edd90e185fde7d9023f683ff3b0345776/examples/ScpTo.java#L43-L45 show you should first grab an InputStream and then connect.

Our code had:
```java
channel.connect();
InputStream in = channel.getInputStream();
List<String> response = readStream(in);
```

It mostly worked, but would occasionally fail, mostly when testing connections between two fastest machines located nearby.

Would it be possible to log a warning when you `getInputStream()` but the channel has already connected, or is that a valid scenario in some circumstances?

norrisjeremy:
Hi @pshem,

I've included a change to add some logger warnings when `getInputStream()` or `getExtInputStream()` are called after `connect()` as part of #431.

Thanks,
Jeremy

pshem:
Thank you very much - my older self would have noticed (we'd spent a week looking over debug level messages from JSch to figure out what was wrong before reading the documentation :smiling_face_with_tear:  )

PR #431, opened 2023-11-13
0.2.13 changes

norrisjeremy:
- #411 add flush operation from is/jsch#39, with new config option to allow disabling in case it causes regressions.
- Update dependencies.
- Add Java 21 to CI pipeline and also use it by default for building.
- Switch CI pipelines to using maven wrapper and produce any execution error messages.
- #403 add a warning when `Channel.getInputStream()` or `Channel.getExtInputStream()` is called after `Channel.connect()`.


norrisjeremy:
FYI, may need to adjust checks for PRs to also mark the Java 21 builds as required after merging this.

Issue #457, opened 2023-12-18
Is jsch affected by the Terrapin attack?


BernhardGruen:
I would like to know if jsch is affected by the Terrapin attack (https://terrapin-attack.com/).
From what I read about the attack I assume jsch is possibly vulnerable to it (if the right ciphers are used).

Thank you for your fork of jsch.

BernhardGruen:
Hey @norrisjeremy 

wow what a fast response - incredible.

Thank you so much for your answer. It seems the corresponding CVE https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-48795 does not yet mention jsch. Maybe you can add jsch to it after the fix is ready.

robtimus:
Snyk also doesn't recognize 0.2.15 as having fixed this: https://security.snyk.io/vuln/SNYK-JAVA-COMGITHUBMWIEDE-6130900

mwiede:
> Snyk also doesn't recognize 0.2.15 as having fixed this: https://security.snyk.io/vuln/SNYK-JAVA-COMGITHUBMWIEDE-6130900

Snyk does not know 0.2.15, latest version in list is 0.2.14

mwiede:
CVE-2023-48795 is fixed in version 0.2.15 or higher.

mohanchavata12:
Which versions of the jsch is vulnerable ? No indication of the Terrapin CVE on the component versions here 
https://mvnrepository.com/artifact/com.jcraft/jsch

norrisjeremy:
Hi @mohanchavata12,

That is a link to the original version of JSch that has been abandoned by the original developer.
The link for this fork of JSch is: https://mvnrepository.com/artifact/com.github.mwiede/jsch.

Thanks,
Jeremy

mohanchavata12:
Thank you Jeremy for your response. As per the NVD, jsch before 0.2.15 is vulnerable to the Terrapin. There is no clear sign of the CVE 2023-48795 here- https://mvnrepository.com/artifact/com.github.mwiede/jsch or either in the Sync. Can I believe jsch before 0.2.15 is vulnerable as per the NVD details and update the component to the 0.2.15 version(fix is available) ? I appreciate your response. Thank you.

PR #461, opened 2023-12-18
0.2.15 changes

norrisjeremy:
- Adjust maven workflow to split unit & integrations into separate steps so that they can be executed via direct mojo calls and thus avoiding potentially reexecuting prior phases.
- Integrate Forbidden API Checker and resolve issues it flagged.
- #457 address CVE-2023-48795 by adding support for new strict key exchange extension.
-- This change introduces two new config options to control usage of the new strict key exchange extension:
--- `enable_strict_kex` (set to `yes` by default)
--- `require_strict_kex` (set to `no` by default)
-- If either option (or both) is enabled, then JSch will attempt to use the new strict key exchange extension.
--- If the `require_strict_kex` option is enabled and JSch detects the server does not support it, then JSch will terminate the connection and throw an exception.
--- If the `require_strict_kex` option is not enabled and JSch detects the server does not support it, then JSch will fallback and proceed with the connection without using the new extension.
--- This gives users the ability to enable a strong security posture if needed and avoid proceeding with connections to potentially insecure servers.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -34,7 +34,7 @@ jobs:
         java-version: '21'
         check-latest: true
     - name: Build with Maven
-      run: ./mvnw -B -V -e -DskipTests=true package
+      run: ./mvnw -B -V -e -DskipTests=true verify
     - uses: actions/upload-artifact@v3
       with:
         name: java-${{ matrix.java }}-jars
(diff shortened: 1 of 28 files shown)


norrisjeremy:
FYI, I can add more extensive unit tests once OpenSSH 9.6 is available in Alpine Linux so that we can use it as a TestContainer.
I have locally built OpenSSH 9.6 and executed my preliminary integration test against it and it appears to work.

mwiede:
Very good, the https://github.com/RUB-NDS/Terrapin-Scanner)[https://github.com/RUB-NDS/Terrapin-Scanner/releases/tag/v1.0.2 reports the following for connecting from the new code:

```
$ ./Terrapin_Scanner_Linux_amd64 --listen 0.0.0.0:2222
Listening for incoming client connection on 0.0.0.0:2222
================================================================================
==================================== Report ====================================
================================================================================

Remote Banner: SSH-2.0-JSCH_0.2.15_SNAPSHOT

ChaCha20-Poly1305 support:   false
CBC-EtM support:             false

Strict key exchange support: true

==> The scanned peer supports Terrapin mitigations and can establish
    connections that are NOT VULNERABLE to Terrapin. Glad to see this.
    For strict key exchange to take effect, both peers must support it.

Note: This tool is provided as is, with no warranty whatsoever. It determines
      the vulnerability of a peer by checking the supported algorithms and
      support for strict key exchange. It may falsely claim a peer to be
      vulnerable if the vendor supports countermeasures other than strict key
      exchange.

For more details visit our website available at https://terrapin-attack.com
```

mwiede (approved):

BernhardGruen:
Much appreciated thank you for your fast fix @norrisjeremy

> FYI, I can add more extensive unit tests once OpenSSH 9.6 is available in Alpine Linux so that we can use it as a TestContainer. I have locally built OpenSSH 9.6 and executed my preliminary integration test against it and it appears to work.

Alpine Linux also just released OpenSSH 9.6 for edge. The 3.19 branch got the upgrade too.

pacion:
Can you please provide me supression for owasp xml for the vulnerability CVE-2023-48795? @norrisjeremy

norrisjeremy:
@pacion No thank you.

PR #463, opened 2023-12-20
Do not falsely log support for ext-info if the server did not return 'ext-info-s'

mvegter:
Opted for not logging, another option would be to log something along the lines of `not supported`


norrisjeremy (approved):
Oof, apologies for missing that.

PR #479, opened 2024-01-14
ChannelSftp cannot download directories, it results in a zero byte fi…

mvegter:
…le, and should thus fail consistently

The file retrievel with a user provided destination as String already validates if the source file is a directory and fails accordingly. This change is for the user provided OutputStream flow where this validation was missing.

### DIFF
--- a/src/test/java/com/jcraft/jsch/SftpRetrievalIT.java
+++ b/src/test/java/com/jcraft/jsch/SftpRetrievalIT.java
@@ -1,0 +1,133 @@
+package com.jcraft.jsch;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import java.io.FileOutputStream;
+import java.io.OutputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Base64;
+import java.util.List;
+import java.util.Locale;
+import java.util.Random;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.images.builder.ImageFromDockerfile;
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+@Testcontainers
+class SftpRetrievalIT {
+
+  private static final int timeout = 10000;
+
+  @TempDir
+  public Path tmpDir;
+  private Path in;
+  private Path out;
+
+  @Container
+  public GenericContainer<?> sshd = new GenericContainer<>(
+      new ImageFromDockerfile().withFileFromClasspath("asyncsshd.py", "docker/asyncsshd.py")
+          .withFileFromClasspath("ssh_host_ed448_key", "docker/ssh_host_ed448_key")
+          .withFileFromClasspath("ssh_host_ed448_key.pub", "docker/ssh_host_ed448_key.pub")
+          .withFileFromClasspath("ssh_host_rsa_key", "docker/ssh_host_rsa_key")
+          .withFileFromClasspath("ssh_host_rsa_key.pub", "docker/ssh_host_rsa_key.pub")
+          .withFileFromClasspath("ssh_host_ecdsa256_key", "docker/ssh_host_ecdsa256_key")
+          .withFileFromClasspath("ssh_host_ecdsa256_key.pub", "docker/ssh_host_ecdsa256_key.pub")
+          .withFileFromClasspath("ssh_host_ecdsa384_key", "docker/ssh_host_ecdsa384_key")
+          .withFileFromClasspath("ssh_host_ecdsa384_key.pub", "docker/ssh_host_ecdsa384_key.pub")
+          .withFileFromClasspath("ssh_host_ecdsa521_key", "docker/ssh_host_ecdsa521_key")
+          .withFileFromClasspath("ssh_host_ecdsa521_key.pub", "docker/ssh_host_ecdsa521_key.pub")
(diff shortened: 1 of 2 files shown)


src/test/java/com/jcraft/jsch/SftpRetrievalIT.java
```diff
@@ -0,0 +1,138 @@
...
+
+import java.io.FileOutputStream;
+import java.io.OutputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Base64;
+import java.util.List;
+import java.util.Locale;
+import java.util.Random;
```
norrisjeremy, line 20: Google Java Format states that all imports (both java.* & non-java.*) should be organized together into a single block alphabetically.

src/test/java/com/jcraft/jsch/SftpRetrievalIT.java
```diff
@@ -0,0 +1,138 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static java.nio.charset.StandardCharsets.UTF_8;
```
norrisjeremy, line 24: Google Java Format states that all static imports should be organized into a block prior to regular imports.

src/test/java/com/jcraft/jsch/SftpRetrievalIT.java
```diff
@@ -0,0 +1,138 @@
...
+
+@Testcontainers
+class SftpRetrievalIT {
+
+  private static final int timeout = 10000;
+
+
```
norrisjeremy, line 31: Extraneous blank line.

src/test/java/com/jcraft/jsch/SftpRetrievalIT.java
```diff
@@ -0,0 +1,138 @@
...
+          .withFileFromClasspath("ssh_host_ecdsa521_key.pub", "docker/ssh_host_ecdsa521_key.pub")
+          .withFileFromClasspath("ssh_host_ed25519_key", "docker/ssh_host_ed25519_key")
+          .withFileFromClasspath("ssh_host_ed25519_key.pub", "docker/ssh_host_ed25519_key.pub")
+          .withFileFromClasspath("ssh_host_dsa_key", "docker/ssh_host_dsa_key")
+          .withFileFromClasspath("ssh_host_dsa_key.pub", "docker/ssh_host_dsa_key.pub")
+          .withFileFromClasspath("authorized_keys", "docker/authorized_keys")
+          .withFileFromClasspath("Dockerfile", "docker/Dockerfile.asyncssh"))
+      .withExposedPorts(22);
+
+
```
norrisjeremy, line 53: Extraneous blank line.

src/test/java/com/jcraft/jsch/SftpRetrievalIT.java
```diff
@@ -0,0 +1,138 @@
...
+  @TempDir
+  public Path tmpDir;
+  private Path in;
+  private Path out;
```
norrisjeremy, line 57: I would put these before the `sshd` above to better match organization of other ITs.

src/test/java/com/jcraft/jsch/SftpRetrievalIT.java
```diff
@@ -0,0 +1,138 @@
...
+    session.connect();
+    ChannelSftp sftp = (ChannelSftp) session.openChannel("sftp");
+    sftp.connect(timeout);
+    sftp.put(in.toString(), "/root/test");
+    method.accept(sftp);
+    sftp.disconnect();
+    session.disconnect();
+  }
+
+
```
norrisjeremy, line 132: Extraneous blank line.

src/test/java/com/jcraft/jsch/SftpRetrievalIT.java
```diff
@@ -0,0 +1,138 @@
...
+  @FunctionalInterface
+  public interface ThrowingConsumer<T, E extends Exception> {
+    void accept(T t) throws E;
+  }
+
```
norrisjeremy, line 137: Extraneous blank line.

norrisjeremy (changes requested):

src/test/java/com/jcraft/jsch/SftpRetrievalIT.java
```diff
@@ -0,0 +1,133 @@
+package com.jcraft.jsch;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static java.nio.charset.StandardCharsets.UTF_8;
```
norrisjeremy, line 5: The static imports should be organized alphabetically.

norrisjeremy (changes requested):

norrisjeremy (commented):
I'm not a huge fan of the nested lambda expressions here as I think they sort of obfuscate the intention of the tests.

I think it would be more readily apparent what the intention of the test is if they simply were `SftpException sftpException = assertThrows(SftpException.class, () -> sftp.get(...));` calls.

You could probably achieve that by changing the `doSftp()` method into a `createSftp()` method that simply did everything up to `sftp.put()` line and then just returns the `sftp` object.

That would also allow you to speed-up the test a bit, as we could drop the `beforeEach()` setup that creates the random data file since it isn't actually needed for anything.

norrisjeremy:
So after reviewing this more closely, I have some reservations about landing this.

The `get()` method with an `OutputStream dst` seems like it may be more intended as more of a "no seatbelts provided" version in that it does not perform validations of the `src` argument in contrast to the version of `get()` with a `String dst`.
Additionally, this change would slightly reduce performance of this method since it will now have to always perform a `_stat()` call even if no `SftpProgressMonitor` is provided, which implies an extra client-to-server message exchange.

It's also possible that somewhere out there, some JSch user is relying upon the existing implementation that does not throw an exception when calling this method for a remote directory source (as unusual as that may seem to us), and by implementing this change we would be suddenly breaking their application.

Considering all of these things, there seems to be little upside to landing this.

Issue #42, opened 2021-04-22
"no passphrase set." for already decrypted IdentityFile (KeyPairDeferred)


adnklauser:
I'm using this fork of jsch together with Spring Integration SFTP. The Spring Integration SFTP wrapper will do `setPassphrase` on an identity file ahead of time. That means I get a `KeyPairDeferred` where `encrypted` is set to `true`, but on its `delegate`, it's set to `false` (as it has been decrypted already by the earlier call to `setPassphrase`)

![image](https://user-images.githubusercontent.com/34234988/115748691-b3174280-a396-11eb-9ba0-641679f6e971.png)

Now `com.jcraft.jsch.UserAuthPublicKey#decryptKey` sees (correctly) that the identity is not encrypted and tries to call `setPassphrase(null)`
```java
if(!identity.isEncrypted() || passphrase!=null){
  if(identity.setPassphrase(passphrase)){
    if(passphrase!=null &&
       (session.getIdentityRepository() instanceof IdentityRepository.Wrapper)){
      ((IdentityRepository.Wrapper)session.getIdentityRepository()).check();
    }
    break;
  }
}
```

But then two calls down in `com.jcraft.jsch.KeyPair#decrypt(byte[])`, we check the `encrypted` field on the `KeyPairDeferred` itself instead, which means that the `decrypt` call doesn't abort early (already decrypted case), but instead falls into the "no passphrase set." case
```java
@Override
public boolean decrypt(byte[] _passphrase) {
    try {
        if (!encrypted) {
            return true;
        }
        if (_passphrase == null) {
            JSch.getLogger().log(Logger.ERROR, "no passphrase set.");
            return false;
        }
```

Shouldn't this be
```java
if(!isEncrypted()) {
```
instead for this scenario to work?

mwiede:
Hi @adnklauser,
Thanks for reporting your issue.
So do I understand correct, you are using an encrypted Openssh-Format Key, then `Identity.setPassphrase` will decrypt it, and then later there is a problem with `KeyPairDeferred` ?
We have a junit Test `KeyPairIT`, which tests both cases, encrypted and unencrypted. Can you please provide a failing unit test for your issue?

mwiede:
Thanks @chklauser I have now merged your testcase and adopted the code. The is no usage of com.jcraft.jsch.KeyPair#decrypt(java.lang.String) inside of the project, but I also overwrote it in KeyPairDeferred to be consistent.

mwiede:
[...]

PR #503, opened 2024-02-15
JSCH pins carrier threads when using virtual threads

bcluap:
Virtual threads pin to carrier threads when a synchronized block has blocking code in it. We've come across 2 major occurrences of this so far. Using explicit locks is the way to resolve this.

### DIFF
--- a/src/main/java/com/jcraft/jsch/Session.java
+++ b/src/main/java/com/jcraft/jsch/Session.java
@@ -40,6 +40,8 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Properties;
 import java.util.Vector;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import javax.crypto.AEADBadTagException;
 
 public class Session {
@@ -128,7 +130,7 @@ public class Session {
   private volatile boolean isAuthed = false;
 
   private Thread connectThread = null;
-  private Object lock = new Object();
+  private final Lock lock = new ReentrantLock();
 
   boolean x11_forwarding = false;
   boolean agent_forwarding = false;
@@ -526,7 +528,8 @@ public class Session {
 
       isAuthed = true;
 
-      synchronized (lock) {
+      lock.lock();
+      try {
         if (isConnected) {
           connectThread = new Thread(this::run);
           connectThread.setName("Connect thread " + host + " session");
@@ -540,6 +543,8 @@ public class Session {
           // The session has been already down and
           // we don't have to start new thread.
         }
+      } finally {
+        lock.unlock();
       }
     } catch (Exception e) {
       in_kex = false;
@@ -1743,7 +1748,8 @@ public class Session {
     boolean require_strict_kex = this.require_strict_kex;
     boolean resetSeqo = packet.buffer.getCommand() == SSH_MSG_NEWKEYS && doStrictKex;
 
-    synchronized (lock) {
+    lock.lock();
+    try {
       encode(packet);
       if (io != null) {
         io.put(packet);
@@ -1754,6 +1760,8 @@ public class Session {
           seqo = 0;
         }
       }
+    } finally {
+      lock.unlock();
     }
 
     if (resetSeqo && io != null && getLogger().isEnabled(Logger.INFO)) {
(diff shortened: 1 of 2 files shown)


norrisjeremy:
I object to this change.
JSch uses traditional synchronized type locking in numerous places, and changing only this one location seems pointless.

Additionally, the entire issue with carrier thread pinning is going to be fixed in a future Java release (see https://mail.openjdk.org/pipermail/loom-dev/2024-February/006433.html), thus making wholesale changes like this unnecessary.

norrisjeremy:
Also, JSch creates one or more traditional (non-virtual) threads per connection (including one of the code blocks you suggested to change to a ReentrantLock), so trying to use it in the context of a virtual thread seems questionable to me anyway.

bcluap:
Hi. I appreciate your sentiments on this. I just changed to a lock instead of the synchronized block around the lock object as it was the simplest way of making the change with the least impact. I was hoping to use virtual threads with jsch before 23 or 24 but for now will wrap calls to it with with a platform thread

norrisjeremy:
Yes, I simply do not understand why Oracle chose to release Virtual threads as a non-preview feature in it's current state: the carrier thread pinning is significant impediment to safely using it for many (perhaps most?) folks, especially when interfacing with legacy code bases.

I also would like to ask: did you __actually__ encounter deadlocks that could be attributed to the carrier thread pinning that could happen when virtual threads are executing the synchronized portions of the JSch code that you proposed we change?

bcluap:
> I also would like to ask: did you **actually** encounter deadlocks that could be attributed to the carrier thread pinning that could happen when virtual threads are executing the synchronized portions of the JSch code that you proposed we change?

Yes - entire JVM hangs and is completely non responsive. We have worked through these and have got to a point where currently we have no pinning. Basically we solved it in 1 of 2 ways:
1) Pass the call to the third party lib onto a platform thread pool
2) Do a PR to upstream project
3) Use an alternative library

The pinning issues we came across thus far are from dependencies we have added (we use Quarkus and its been pretty good at addressing pinning in its dependencies) :
- Jetty websocket client
- Hazelcast startup and shutdown API's
- Apache commons email
- AWS KMS sdk
- Azure document analysis sdk
- AWS recognition sdk
- Jsch
- Apache CXF
- MySQL driver (We changed to MariaDB driver and it does not pin)

JSCH pinning:
2024-02-15 11:00:04.106,"Thread[#42,ForkJoinPool-1-worker-2,5,CarrierThreads]"
2024-02-15 11:00:04.106,    java.base/java.lang.VirtualThread$VThreadContinuation.onPinned(VirtualThread.java:183)
2024-02-15 11:00:04.106,    java.base/jdk.internal.vm.Continuation.onPinned0(Continuation.java:393)
2024-02-15 11:00:04.106,    java.base/java.lang.VirtualThread.tryYield(VirtualThread.java:756)
2024-02-15 11:00:04.106,    java.base/java.lang.Thread.yield(Thread.java:443)
2024-02-15 11:00:04.106,    com.jcraft.jsch.Session.disconnect(Session.java:2114) <== monitors:1

2024-02-15 11:00:04.025,    java.base/sun.nio.ch.NioSocketImpl$1.read(NioSocketImpl.java:796)
2024-02-15 11:00:04.025,    java.base/java.net.Socket$SocketInputStream.read(Socket.java:1099)
2024-02-15 11:00:04.025,    com.jcraft.jsch.IO.getByte(IO.java:93)
2024-02-15 11:00:04.025,    com.jcraft.jsch.Session.read(Session.java:1183)
2024-02-15 11:00:04.025,    com.jcraft.jsch.UserAuthPublicKey._start(UserAuthPublicKey.java:209)
2024-02-15 11:00:04.025,    com.jcraft.jsch.UserAuthPublicKey.start(UserAuthPublicKey.java:114) <== monitors:1
2024-02-15 11:00:04.025,    com.jcraft.jsch.Session.connect(Session.java:480)

norrisjeremy:
This only serves to prove my point that Virtual threads are simply not ready for prime-time and should not have been released as a non-preview feature in the first place until Oracle actually solves the carrier thread pinning issue within the JDK itself.

norrisjeremy:
FYI, it appears Oracle is making progress on solving the carrier thread pinning issue, see https://mail.openjdk.org/pipermail/loom-dev/2024-May/006632.html.

norrisjeremy:
FYI, it appears [JEP-491](https://openjdk.org/jeps/491) will be included as part of the OpenJDK 24 release, which should largely solve the issue with carrier thread pinning due to use of `synchronize`.
As such, I believe this PR can be closed now, as it should be unnecessary.

PR #529, opened 2024-03-28
support get new format fingerprint from HostKey

eshizhan:
The PR make `HostKey.getFingerPrint` method support new format fingerprint, and keeping original method return old format.

### DIFF
--- a/src/main/java/com/jcraft/jsch/HostKey.java
+++ b/src/main/java/com/jcraft/jsch/HostKey.java
@@ -118,6 +118,10 @@ public class HostKey {
   }
 
   public String getFingerPrint(JSch jsch) {
+    return getFingerPrint(jsch, false);
+  }
+
+  public String getFingerPrint(JSch jsch, boolean isNewFormat) {
     HASH hash = null;
     try {
       String _c = JSch.getConfig("FingerprintHash").toLowerCase(Locale.ROOT);
@@ -128,7 +132,7 @@ public class HostKey {
         jsch.getInstanceLogger().log(Logger.ERROR, "getFingerPrint: " + e.getMessage(), e);
       }
     }
-    return Util.getFingerPrint(hash, key, false, true);
+    return Util.getFingerPrint(hash, key, isNewFormat, !isNewFormat);
   }
 
   public String getComment() {
--- a/src/test/java/com/jcraft/jsch/KnownHostsTest.java
+++ b/src/test/java/com/jcraft/jsch/KnownHostsTest.java
@@ -1057,6 +1057,8 @@ class KnownHostsTest {
     assertEquals(
         "9d:38:5b:83:a9:17:52:92:56:1a:5e:c4:d4:81:8e:0a:ca:51:a2:64:f1:74:20:11:2e:f8:8a:c3:a1:39:49:8f",
         key.getFingerPrint(jsch), "fingerprint mismatch");
+    assertEquals("SHA256:nThbg6kXUpJWGl7E1IGOCspRomTxdCARLviKw6E5SY8",
+        key.getFingerPrint(jsch, true), "fingerprint mismatch");
     assertEquals(expectedHostResult, key.getHost(), "host mismatch");
     assertEquals(rsaKey, key.getKey(), "key mismatch");
     assertEquals(expectedMarker, key.getMarker(), "marker mismatch");


norrisjeremy:
Hi @eshizhan,

What exactly do you mean by "new format fingerprint"?

Thanks,
Jeremy

eshizhan:
The old fingerprint format like: `9d:38:5b:83:a9:17:52:92:56:1a:5e:c4:d4:81:8e:0a:ca:51:a2:64:f1:74:20:11:2e:f8:8a:c3:a1:39:49:8f`, the new fingerprint format like: `SHA256:nThbg6kXUpJWGl7E1IGOCspRomTxdCARLviKw6E5SY8`.
The `Util.getFingerPrint` in original code is support two format, just `HostKey.getFingerPrint` using old format before.

norrisjeremy:
Hi @eshizhan,

I believe the only effect that your change has is to force the fingerprint has to prepend the `MD5:` or `SHA256:` string to the front of the string hash that it returns (depending upon which algorithm is configured): otherwise nothing else is different.

I'm not really sure that is thus worthwhile to introduce an unwieldy method overload (with a confusing `boolean newFormat` argument) to the public API.

Can you provide more details as to how this change will solve a functional issue you are having?

Thanks,
Jeremy

norrisjeremy:
Hi @eshizhan,

What would you think of something like this instead, to just force `HostKey.getFingerPrint()` into always using the more modern format introduced in OpenSSH 6.8?

```diff
diff --git a/src/main/java/com/jcraft/jsch/HostKey.java b/src/main/java/com/jcraft/jsch/HostKey.java
index 0f9922b..112478a 100644
--- a/src/main/java/com/jcraft/jsch/HostKey.java
+++ b/src/main/java/com/jcraft/jsch/HostKey.java
@@ -128,7 +128,7 @@ public class HostKey {
         jsch.getInstanceLogger().log(Logger.ERROR, "getFingerPrint: " + e.getMessage(), e);
       }
     }
-    return Util.getFingerPrint(hash, key, false, true);
+    return Util.getFingerPrint(hash, key, true, false);
   }
 
   public String getComment() {
diff --git a/src/test/java/com/jcraft/jsch/KnownHostsTest.java b/src/test/java/com/jcraft/jsch/KnownHostsTest.java
index c74df31..2eac7e0 100644
--- a/src/test/java/com/jcraft/jsch/KnownHostsTest.java
+++ b/src/test/java/com/jcraft/jsch/KnownHostsTest.java
@@ -454,8 +454,7 @@ class KnownHostsTest {
         "|1|AAECAwQFBgcICQoLDA0ODxAREhMUFRYXGBkaGxwdHh8=|mie6rcAf1aPGk6d+HxnkpvO4HaOAH/Y6YWegs+Xog/s=",
         hhk.getHost(), "host mismatch");
     assertEquals("", hhk.getMarker(), "marker mismatch");
-    assertEquals(
-        "9c:fb:7f:99:79:01:6d:46:68:87:39:15:4f:f5:cc:9d:71:7a:8b:5a:4a:c1:c7:4b:9c:20:a5:91:c2:6a:ff:5a",
+    assertEquals("SHA256:nPt/mXkBbUZohzkVT/XMnXF6i1pKwcdLnCClkcJq/1o",
         hhk.getFingerPrint(jsch));
     assertEquals(null, hhk.getComment(), "comment mismatch");
     assertEquals("ICAgIHNzaC1yc2E=", hhk.getKey(), "key mismatch");
@@ -1006,8 +1005,7 @@ class KnownHostsTest {
       String expectedType, String expectedComment) {
     assertEquals(expctedHost, hhk.getHost(), "host mismatch");
     assertEquals(expectedMarker, hhk.getMarker(), "marker mismatch");
-    assertEquals(
-        "1e:b5:70:92:65:6e:6a:f9:d6:7a:a9:43:00:40:a2:e7:c8:51:35:df:ee:60:19:b7:4b:18:1d:eb:46:48:28:4b",
+    assertEquals("SHA256:HrVwkmVuavnWeqlDAECi58hRNd/uYBm3Sxgd60ZIKEs",
         hhk.getFingerPrint(jsch));
     assertEquals(expectedComment, hhk.getComment(), "comment mismatch");
     assertEquals("ICAgIHNzaC1kc2E=", hhk.getKey(), "key mismatch");
@@ -1022,8 +1020,7 @@ class KnownHostsTest {
     assertEquals("|1|AAECAwQFBgcICQoLDA0ODxAREhM=|/pE4peaossRYDRp6bEWa348eFLI=", hhk.getHost(),
         "host mismatch");
     assertEquals(expectedMarker, hhk.getMarker(), "marker mismatch");
-    assertEquals(
-        "1e:b5:70:92:65:6e:6a:f9:d6:7a:a9:43:00:40:a2:e7:c8:51:35:df:ee:60:19:b7:4b:18:1d:eb:46:48:28:4b",
+    assertEquals("SHA256:HrVwkmVuavnWeqlDAECi58hRNd/uYBm3Sxgd60ZIKEs",
         hhk.getFingerPrint(jsch));
     assertEquals(expectedComment, hhk.getComment(), "comment mismatch");
     assertEquals("ICAgIHNzaC1kc2E=", hhk.getKey(), "key mismatch");
@@ -1054,8 +1051,7 @@ class KnownHostsTest {
     assertEquals(1, keys.length, "1 key expected");
     HostKey key = keys[0];
     assertEquals("some comment", key.getComment(), "comment mismatch");
-    assertEquals(
-        "9d:38:5b:83:a9:17:52:92:56:1a:5e:c4:d4:81:8e:0a:ca:51:a2:64:f1:74:20:11:2e:f8:8a:c3:a1:39:49:8f",
+    assertEquals("SHA256:nThbg6kXUpJWGl7E1IGOCspRomTxdCARLviKw6E5SY8",
         key.getFingerPrint(jsch), "fingerprint mismatch");
     assertEquals(expectedHostResult, key.getHost(), "host mismatch");
     assertEquals(rsaKey, key.getKey(), "key mismatch");
```

Thanks,
Jeremy

eshizhan:
@norrisjeremy Thank you for the review.
I have considered two use cases: 1. Comparing whether the HostKey hash value is correct. 2. Logging to help with auditing and troubleshooting. These are also the functions of the fingerprint. 

Initially, I had the same idea as you, to enforce the use of the new fingerprint format. But considering that this library is a well-known library in Java and is used by many other libraries and applications, I worry that forcing the transition to the new format will cause incompatibility. For example, someone might persist the HostKey hash to be compared in the old format.

This new format has actually been released for a long time. Switching to the new format is definitely a good choice. I chose the clumsy way of overloading, simply because of concerns about compatibility.

DukeAstar:
I think the merged coud lead to regression if value is used as compare value.

Another option is to let user decide which format it prefers, while keeping new format as default.
This way users can keep access to old format if needed and it will be more flexible

See below my proposal.
@mwiede @norrisjeremy   : do you think i can make a pull request ?

`
public String getFingerPrint(JSch jsch) {
		    return getFingerPrint(jsch, true, false);
		  }
	
	  public String getFingerPrint(JSch jsch, boolean include_prefix, boolean force_hex) {
	    HASH hash = null;
	    try {
	      String _c = JSch.getConfig("FingerprintHash").toLowerCase();
	      Class<? extends HASH> c = Class.forName(JSch.getConfig(_c)).asSubclass(HASH.class);
	      hash = c.getDeclaredConstructor().newInstance();
	    } catch (Exception e) {
	      if (jsch.getInstanceLogger().isEnabled(Logger.ERROR)) {
	        jsch.getInstanceLogger().log(Logger.ERROR, "getFingerPrint: " + e.getMessage(), e);
	      }
	    }
	    return Util.getFingerPrint(hash, key, include_prefix, force_hex);
	  }`

norrisjeremy:
Hi @DukeAstar,

No, I would not approve a pull request like that.
The new format was introduced by OpenSSH over 9 years ago, so I deliberately made this change to catch up with them.

Thanks,
Jeremy

PR #622, opened 2024-08-16
Add logging stack trace on auth failure

dbeliakov-r7:
When authentication fails, logger only saves the exception message. This is often not enough to understand what caused the failure.
This change adds the stack trace to the error message.

### DIFF
--- a/src/main/java/com/jcraft/jsch/Session.java
+++ b/src/main/java/com/jcraft/jsch/Session.java
@@ -500,8 +500,8 @@ public class Session {
               // SSH_MSG_DISCONNECT: Too many authentication failures
               // System.err.println("ee: " + ee);
               if (getLogger().isEnabled(Logger.WARN)) {
-                getLogger().log(Logger.WARN,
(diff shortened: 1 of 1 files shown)


norrisjeremy (changes requested):
No, this is not the way to implement this.
You can simply add the exception as a third argument to the log method call.
I.e `getLogger.log(Logger.WARN, ..., ee)`.

norrisjeremy:
> No, this is not the way to implement this. You can simply add the exception as a third argument to the log method call. I.e `getLogger.log(Logger.WARN, ..., ee)`.

See https://github.com/mwiede/jsch/blob/d064f44b5ae4c9447908826df8f5f5eba7ee5a84/src/main/java/com/jcraft/jsch/Session.java#L1443 for an example how to log the exception stack trace correctly.

norrisjeremy:
Can you squash this into a single commit?

norrisjeremy (approved):

PR #623, opened 2024-08-16
Single authentication attempt and time out.

dbeliakov-r7:
When authenticating with a script which uses the same password every time, the consecutive attempts with the incorrect password give the same result.
This change adds a preference for a single authentication attempt. When enabled, authentication ends after the first failure and doesn't ask to retry.

Additionally, when time out is set in the session, authentication shouldn't take longer than this time out.

### DIFF
--- a/src/main/java/com/jcraft/jsch/UserAuthKeyboardInteractive.java
+++ b/src/main/java/com/jcraft/jsch/UserAuthKeyboardInteractive.java
@@ -72,7 +72,15 @@ class UserAuthKeyboardInteractive extends UserAuth {
       session.write(packet);
 
       boolean firsttime = true;
+      final long timeout = System.currentTimeMillis() + session.getTimeout();
       loop: while (true) {
+        if (session.getTimeout() > 0 && System.currentTimeMillis() > timeout) {
+          throw new JSchAuthCancelException("keyboard-interactive");
+        }
+        if (session.auth_failures >= session.max_auth_tries) {
+          return false;
+        }
+
         buf = session.read(buf);
         int command = buf.getCommand() & 0xff;
 


norrisjeremy:
I don't fully grasp what this change is attempting to fix: can you provide a more concrete example of what the issue is and how this solves it?

src/main/java/com/jcraft/jsch/UserAuthKeyboardInteractive.java
```diff
@@ -193,4 +204,9 @@ public boolean start(Session session) throws Exception {
     }
     // return false;
   }
+
+  private boolean isMaxFailuresReached(int initial, int current) {
+    return Boolean.parseBoolean(JSch.getConfig("user_auth_keyboard_interactive_single_attempt"))
```
norrisjeremy, line 209: This needs to use `session.getConfig()`, not `JSch.getConfig()` so that it can be set on a per Session basis.

I believe you also need to add the config item to `Session.applyConfig()` via a `Session.checkConfig()` call.
See https://github.com/mwiede/jsch/blob/d064f44b5ae4c9447908826df8f5f5eba7ee5a84/src/main/java/com/jcraft/jsch/Session.java#L3207.

src/main/java/com/jcraft/jsch/UserAuthPassword.java
```diff
@@ -162,6 +162,13 @@ public boolean start(Session session) throws Exception {
               throw new JSchPartialAuthException(Util.byte2str(foo));
             }
             session.auth_failures++;
+
+            // Do not ask re-entering password after the first failure. Useful when authenticating
+            // with a script that uses the same password every time.
+            if (Boolean.parseBoolean(JSch.getConfig("user_auth_single_attempt"))) {
```
norrisjeremy, line 168: This needs to use `session.getConfig()`, not `JSch.getConfig()` so that it can be set on a per Session basis.

I believe you also need to add the config item to `Session.applyConfig()` via a `Session.checkConfig()` call.
See https://github.com/mwiede/jsch/blob/d064f44b5ae4c9447908826df8f5f5eba7ee5a84/src/main/java/com/jcraft/jsch/Session.java#L3207.

norrisjeremy (changes requested):

dbeliakov-r7:
Thanks for the quick reply and review 👍 
So, the intent is to reduce time spent on auth by removing unnecessary steps. With the time out, I set time out to the session, but the auth could spend more time in that loop that time-out allows. 

Regarding the single attempt, the example is when I use a script and it has an incorrect password I want to fail fast and don't retry because for the next attempt the script will pass the same password and the result will be same, just more time spent.

Actually, now it seems that the config option is redundant and the same could be achieved by 
```
if (session.auth_failures >= session.max_auth_tries) {
          return false;
}
```
as it is done in UserAuthPassword, just need to add that to keyboard interactive and then use `max_auth_retries` to limit retries.

PR #624, opened 2024-08-19
Replace old Hashtable with modern data structures.

dbeliakov-r7:
This PR replaces the old Hashtable with modern data structures. Hashtable is no longer maintained and is not optimised for better performance.

I replaced type declaration with the interface. Since Hashtable is-a Map, this is seamless.
I replaced Hashtable with ConcurrentHashMap where multiple threads is expected to access the map, but ConcurrentHashMap handles it better (this could be improved further, but one step at a time).
I replaced Hashtable with HashMap where synchronisation was unnecessary.

### DIFF
--- a/src/main/java/com/jcraft/jsch/ChannelSession.java
+++ b/src/main/java/com/jcraft/jsch/ChannelSession.java
@@ -26,15 +26,16 @@
 
 package com.jcraft.jsch;
 
-import java.util.Enumeration;
 import java.util.Hashtable;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 class ChannelSession extends Channel {
   private static byte[] _session = Util.str2byte("session");
 
   protected boolean agent_forwarding = false;
   protected boolean xforwading = false;
-  protected Hashtable<byte[], byte[]> env = null;
+  protected Map<byte[], byte[]> env = null;
 
   protected boolean pty = false;
 
@@ -78,7 +79,7 @@ class ChannelSession extends Channel {
   @Deprecated
   public void setEnv(Hashtable<byte[], byte[]> env) {
     synchronized (this) {
-      this.env = env;
+      this.env = new ConcurrentHashMap<>(env);
     }
   }
 
@@ -103,14 +104,11 @@ class ChannelSession extends Channel {
    */
   public void setEnv(byte[] name, byte[] value) {
     synchronized (this) {
-      getEnv().put(name, value);
+      if (env == null) {
+        env = new ConcurrentHashMap<>();
+      }
+      env.put(name, value);
     }
-  }
-
-  private Hashtable<byte[], byte[]> getEnv() {
-    if (env == null)
-      env = new Hashtable<>();
-    return env;
   }
 
   /**
@@ -205,11 +203,9 @@ class ChannelSession extends Channel {
     }
 
     if (env != null) {
-      for (Enumeration<byte[]> _env = env.keys(); _env.hasMoreElements();) {
-        byte[] name = _env.nextElement();
-        byte[] value = env.get(name);
+      for (Map.Entry<byte[], byte[]> e : env.entrySet()) {
         request = new RequestEnv();
-        ((RequestEnv) request).setEnv(toByteArray(name), toByteArray(value));
+        ((RequestEnv) request).setEnv(toByteArray(e.getKey()), toByteArray(e.getValue()));
         request.request(_session, this);
       }
     }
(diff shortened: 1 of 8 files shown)


norrisjeremy:
I'm not sure I approve of this: it's not fixing any sort of bug nor is it likely fixing any serious performance issue.
All it is doing is potentially introducing new subtle bugs or incompatibilities.

src/test/java/com/jcraft/jsch/JSchTest.java
```diff
@@ -3,10 +3,12 @@
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertSame;
 
-import java.util.Hashtable;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
+import java.util.HashMap;
+import java.util.Map;
```
norrisjeremy, line 10: I'm fairly certain this doesn't meet Google Java style that we utilize.
You should run `mvn formatter:format`.

src/main/java/com/jcraft/jsch/JSch.java
```diff
@@ -608,14 +608,10 @@ public static String getConfig(String key) {
...
+  public static void setConfig(Map<String, String> newconf) {
     synchronized (config) {
-      for (Enumeration<String> e = newconf.keys(); e.hasMoreElements();) {
-        String newkey = e.nextElement();
-        String key =
-            (newkey.equals("PubkeyAcceptedKeyTypes") ? "PubkeyAcceptedAlgorithms" : newkey);
-        config.put(key, newconf.get(newkey));
-      }
+      newconf.forEach((key, value) -> config
+          .put(key.equals("PubkeyAcceptedKeyTypes") ? "PubkeyAcceptedAlgorithms" : key, value));
```
norrisjeremy, line 614: I would rather just see a traditional for loop over an `newconf.entrySet()` instead of the `.forEach()` lambda.

src/main/java/com/jcraft/jsch/JSch.java
```diff
@@ -27,15 +27,15 @@
...
+import java.util.Map;
 import java.util.Vector;
+import java.util.concurrent.ConcurrentHashMap;
 
 public class JSch {
   /** The version number. */
   public static final String VERSION = Version.getVersion();
 
-  static Hashtable<String, String> config = new Hashtable<>();
+  static Map<String, String> config = new ConcurrentHashMap<>();
```
norrisjeremy, line 38: I don't think there is any value in using a `ConcurrentHashMap` instead of a regular `HashMap` since all accesses to the `config` are synchronized on `config` anyway.

src/main/java/com/jcraft/jsch/ChannelX11.java
```diff
@@ -45,8 +46,8 @@ class ChannelX11 extends Channel {
   static byte[] cookie = null;
   private static byte[] cookie_hex = null;
 
-  private static Hashtable<Session, byte[]> faked_cookie_pool = new Hashtable<>();
-  private static Hashtable<Session, byte[]> faked_cookie_hex_pool = new Hashtable<>();
+  private static final Map<Session, byte[]> faked_cookie_pool = new ConcurrentHashMap<>();
+  private static final Map<Session, byte[]> faked_cookie_hex_pool = new ConcurrentHashMap<>();
```
norrisjeremy, line 50: I don't think there is any value in using a ConcurrentHashMap instead of a regular HashMap since all accesses to the faked_cookie_hex_pool are synchronized on faked_cookie_hex_pool anyway.

src/main/java/com/jcraft/jsch/ChannelX11.java
```diff
@@ -45,8 +46,8 @@ class ChannelX11 extends Channel {
   static byte[] cookie = null;
   private static byte[] cookie_hex = null;
 
-  private static Hashtable<Session, byte[]> faked_cookie_pool = new Hashtable<>();
-  private static Hashtable<Session, byte[]> faked_cookie_hex_pool = new Hashtable<>();
+  private static final Map<Session, byte[]> faked_cookie_pool = new ConcurrentHashMap<>();
```
norrisjeremy, line 49: I would just keep faked_cookie_pool a HashMap and maintain the `synchronized` on `faked_cookie_pool` that was removed anyway.

src/main/java/com/jcraft/jsch/ChannelSession.java
```diff
@@ -78,7 +79,7 @@ public void setXForwarding(boolean enable) {
   @Deprecated
   public void setEnv(Hashtable<byte[], byte[]> env) {
     synchronized (this) {
-      this.env = env;
+      this.env = new ConcurrentHashMap<>(env);
```
norrisjeremy, line 82: I don't think there is any value in using a ConcurrentHashMap instead of a regular HashMap since all accesses to the env are synchronized on env anyway.

norrisjeremy (changes requested):

PR #649, opened 2024-09-04
Incorrect Oid for service name in GSSAPI authentication. Fix #648

### LINKED ISSUE #648: Incorrect Oid for service name in GSSAPI authentication
The Oid used to create the service principal name for the target host is incorrect.
This causes the incorrect determination of the realm when the host dns name does not match the realm name.

Description:
Suppose you have two realms EXAMPLE.COM and COMPANY.COM with a trust estabilished between them, and host named "my.example.com" that belongs to the domain COMPANY.COM, with this krb5.conf:
```
[libdefault]
default_realm = EXAMPLE.COM

[domain_realm]
my.example.com = COMPANY.COM
.example.com = EXAMPLE.COM
example.com = EXAMPLE.COM
.company.com = COMPANY.COM
company.com = COMPANY.COM
```

Currently, the SPN generated is incorrect as `host/my.example.com@EXAMPLE.COM`, where it should be `host/my.example.com@COMPANY.COM` instead.

vpinna80:


norrisjeremy (approved):
I'm not especially familiar with GSSAPI authentication.
However, perusing OpenSSH code I believe this change will align with their behavior.

See:
- https://github.com/openssh/openssh-portable/blob/a8ad7a2952111c6ce32949a775df94286550af6b/gss-genr.c#L236
- https://github.com/openssh/openssh-portable/blob/a8ad7a2952111c6ce32949a775df94286550af6b/gss-genr.c#L241

PR #692, opened 2024-10-28
Update Deflate.java

mjmst74:
Updates back to match the logic from jzlib's Deflate class to set len, minus jzlib's unnecessary cast to (int)

`len = MAX_MATCH - (int)(strend - scan);`

All tests are passing.  I did not add a test to the codebase, but with a 105mb XML file, compression level 6

before this update:
- time to deflate: 119.6 seconds
- deflated size: 70mb

after this update:
- time to deflate: 2.4 seconds
- deflated size: 6.2mb

### DIFF
--- a/src/main/java/com/jcraft/jsch/jzlib/Deflate.java
+++ b/src/main/java/com/jcraft/jsch/jzlib/Deflate.java
@@ -1309,7 +1309,9 @@ final class Deflate implements Cloneable {
           && window[++scan] == window[++match] && window[++scan] == window[++match]
           && scan < strend);
 
-      len = MAX_MATCH - strend - scan;
+      // len here is set to MAX_MATCH minus (strend minus scan, i.e. the distance between scan and
+      // strend) to ensure that best_len is set to the best possible length to lookahead
+      len = MAX_MATCH - (strend - scan);
       scan = strend - MAX_MATCH;
 
       if (len > best_len) {


src/main/java/com/jcraft/jsch/jzlib/Deflate.java
```diff
@@ -1309,7 +1309,7 @@ int longest_match(int cur_match) {
           && window[++scan] == window[++match] && window[++scan] == window[++match]
           && scan < strend);
 
-      len = MAX_MATCH - strend - scan;
+      len = MAX_MATCH - (strend - scan);
```
norrisjeremy, line 1312: Can you add a comment here explaining that the parenthesis are required?
Otherwise someone may in the future believe they are unnecessary and remove them.

mjmst74: No problem - I added a comment, based on my best understanding of the longest_match method.

norrisjeremy: Can you please run `mvn formatter:format` so this comment is formatted correctly?

mjmst74: In progress.

norrisjeremy: Can you pull the third line of this comment up to the second line so that it doesn't appear to awkward?
I.e.
```
      // len here is set to MAX_MATCH minus (strend minus scan, i.e. the distance between scan and
      // strend) to ensure that best_len is set to the best possible length to lookahead
```

norrisjeremy (approved):
I believe this should be fine now, thanks!

Issue #739, opened 2025-01-07
SignatureECDSAN destroys private key


ikucuze:
Jsch instance should be reusable.
code as below should work:
```
  JSch jsch = new JSch();
  jsch.addIdentity("id_ecdsa", privateKeyBytes, null, null);
  
  Session session1 = jsch.getSession("username","host", 22);
//...
  Session session2 = jsch.getSession("username","host", 22);
//...
```

jsch.addidentity converts the key to an IdentityFile:
Identity identity=IdentityFile.newInstance(name, prvkey, pubkey, this);

When using a EDCSA key pair, SSH connection goes through signing a payload:
UserAuthPublicKey.start
> Identity.getSignature
>> KeyPairECDSA.getSignature

The impl in KeyPairECDSA is:
```
      SignatureECDSA ecdsa = c.getDeclaredConstructor().newInstance();
      ecdsa.init();
      ecdsa.setPrvKey(prv_array);

      ecdsa.update(data);
      byte[] sig = ecdsa.sign();
```

prv_array contains the in-memory reference to the private key. It is passed to the signature class to perform to signing.

SignatureECDSAN implements that:
```
    // d must be unsigned value.
    d = insert0(d);
```

and insert0 check if the first byte of the private key is positive. If not it makes a copy of the private key with an addition leading zero to make it positive when converted to a number.
```
  private static byte[] insert0(byte[] buf) {
    if ((buf[0] & 0x80) == 0)
      return buf;
    byte[] tmp = new byte[buf.length + 1];
    System.arraycopy(buf, 0, tmp, 1, buf.length);
    Util.bzero(buf);
    return tmp;
  }
```

It finishes with `Util.bzero(buf);`, which clear the parameter array, with is the private key given by keypair, not a copy.

This lead to 0-ing the private key in KeyPairECDSA.

First connection will succeed, second one will fail to provide anymore valid signature du to the 0-ed private key.

---
Workaround:
recreate jsch instance with identity for every connections...

---
fix:
insert0 is called by setPrvKey and setPubKey within SignatureECDSAN
This must not lead to destroying the private / public key
2 ways:
- duplicate every key passed to SignatureECDSAN to prevent destroying it. This leads to possible mistakes later on for future calls... 
- or ensure that the arrays for both these function never get 0-ed

The only other call to this method is :
```
      r = trimLeadingZeros(insert0(r));
      s = trimLeadingZeros(insert0(s));
```
[...]

PR #740, opened 2025-01-07
fix(739) SignatureECDSAN destroying private key

ikucuze:
fix for bug 739

### DIFF
--- a/src/main/java/com/jcraft/jsch/KeyPairECDSA.java
+++ b/src/main/java/com/jcraft/jsch/KeyPairECDSA.java
@@ -345,13 +345,16 @@ class KeyPairECDSA extends KeyPair {
 
   @Override
   public byte[] getSignature(byte[] data) {
+    byte[] keyCopy = null;
     try {
       Class<? extends SignatureECDSA> c =
           Class.forName(JSch.getConfig("ecdsa-sha2-" + Util.byte2str(name)))
               .asSubclass(SignatureECDSA.class);
       SignatureECDSA ecdsa = c.getDeclaredConstructor().newInstance();
       ecdsa.init();
-      ecdsa.setPrvKey(prv_array);
+      // https://github.com/mwiede/jsch/issues/739 : prv_array could be destroyed by ecdsa signing
+      keyCopy = Arrays.copyOf(prv_array, prv_array.length);
+      ecdsa.setPrvKey(keyCopy);
 
       ecdsa.update(data);
       byte[] sig = ecdsa.sign();
@@ -364,6 +367,8 @@ class KeyPairECDSA extends KeyPair {
       if (instLogger.getLogger().isEnabled(Logger.ERROR)) {
         instLogger.getLogger().log(Logger.ERROR, "failed to generate signature", e);
       }
+    } finally {
+      Util.bzero(keyCopy);
     }
     return null;
   }
@@ -390,7 +395,9 @@ class KeyPairECDSA extends KeyPair {
         r_array = tmp[0];
         s_array = tmp[1];
       }
-      ecdsa.setPubKey(r_array, s_array);
+      // https://github.com/mwiede/jsch/issues/739 : keys could be destroyed by ecdsa verification
+      ecdsa.setPubKey(Arrays.copyOf(r_array, r_array.length),
+          Arrays.copyOf(s_array, s_array.length));
       return ecdsa;
     } catch (Exception e) {
       if (instLogger.getLogger().isEnabled(Logger.ERROR)) {


ikucuze:
fixes: https://github.com/mwiede/jsch/issues/739

norrisjeremy:
No, I very much oppose this.
We are not going to stop zero'ing private key data.

ikucuze:
PR updated to always perform a copy of the array instead of not 0-ing it

norrisjeremy (approved):
This seems acceptable to me by best limiting any potential impact.

src/main/java/com/jcraft/jsch/KeyPairECDSA.java
```diff
@@ -351,7 +351,8 @@ public byte[] getSignature(byte[] data) {
               .asSubclass(SignatureECDSA.class);
       SignatureECDSA ecdsa = c.getDeclaredConstructor().newInstance();
       ecdsa.init();
-      ecdsa.setPrvKey(prv_array);
+      // issue 730: prv_array could be destroyed by ecdsa signing process if its first bit is 1
```
norrisjeremy, line 354: Can you fix this to correctly point to issue 739 and not issue 730 (which is unrelated).

norrisjeremy (changes requested):

src/main/java/com/jcraft/jsch/KeyPairECDSA.java
norrisjeremy: Also, I would prefer if we included a full link to `https://github.com/mwiede/jsch/issues/739` in the comment text, so that future readers don't have to figure out what `issue 739` means.

src/main/java/com/jcraft/jsch/KeyPairECDSA.java
```diff
@@ -351,7 +351,8 @@ public byte[] getSignature(byte[] data) {
               .asSubclass(SignatureECDSA.class);
       SignatureECDSA ecdsa = c.getDeclaredConstructor().newInstance();
       ecdsa.init();
-      ecdsa.setPrvKey(prv_array);
+      // issue 739: prv_array could be destroyed by ecdsa signing process if its first bit is 1
```
norrisjeremy, line 354: Can you change this to point at `https://github.com/mwiede/jsch/issues/739` instead of `issue 739`?
That way future readers don't have to figure out how to find what `issue 739` means.

src/main/java/com/jcraft/jsch/KeyPairECDSA.java
ikucuze: maked me realize 2 things:
This copy of the private key has to be cleaned... as we don't know if ecdsa signing will clean it...
So I need Util.bzero to be changed to public.
Would you prefer then:
```
      byte[] keyCopy = Arrays.copyOf(prv_array, prv_array.length);
      ecdsa.setPrvKey(keyCopy);
      Util.bzero(keyCopy);
```


I also need to change the caller of setPubKey with the same fix :(

ikucuze: I pushed a new secured fixed regarding cleaning the key copy.
The resulting code starts to get ugly because of that unneeded copy...

src/main/java/com/jcraft/jsch/KeyPairECDSA.java
ikucuze: done

src/main/java/com/jcraft/jsch/KeyPairECDSA.java
norrisjeremy: > made me realize 2 things: This copy of the private key has to be cleaned... as we don't know if ecdsa signing will clean it... So I need Util.bzero to be changed to public. Would you prefer then:
> 
> ```
>       byte[] keyCopy = Arrays.copyOf(prv_array, prv_array.length);
>       ecdsa.setPrvKey(keyCopy);
>       Util.bzero(keyCopy);
> ```
> 
> I also need to change the caller of setPubKey with the same fix :(

There is zero reason to zero the public key, as that is not sensitive data.

src/main/java/com/jcraft/jsch/jce/Util.java
```diff
@@ -26,8 +26,8 @@
 
 package com.jcraft.jsch.jce;
 
-class Util {
-  static void bzero(byte[] foo) {
+public class Util {
```
norrisjeremy, line 29: There is no reason for this class to be made public.
Please revert this change.

ikucuze:
> `escda.setPrvKey(Arrays.copyOf(prv_array, prv_array.length));`.

This is not acceptable as the copy array might stay in memory without cleaning in case of a positive number, thus setPrvKey won't clean it.
You fought for keeping the cleaning in SignatureECDSAN even if the priv_key is a parameter and not under its responsability. I think the responsability of the cleaning is in the hand of the code originating the data. Thus KeyPairECDSA is now reponsible to clean this copy, and for me was also was responsible for cleaning the original prv_array (and it does that correctly in its dispose() ).

src/main/java/com/jcraft/jsch/jce/Util.java
ikucuze: needed to clean the copy

norrisjeremy:
> > `escda.setPrvKey(Arrays.copyOf(prv_array, prv_array.length));`.
> 
> This is not acceptable as the copy array might stay in memory without cleaning in case of a positive number, thus setPrvKey won't clean it. You fought for keeping the cleaning in SignatureECDSAN even if the priv_key is a parameter and not under its responsability. I think the responsability of the cleaning is in the hand of the code originating the data. Thus KeyPairECDSA is now reponsible to clean this copy, and for me was also was responsible for cleaning the original prv_array (and it does that correctly in its dispose() ).

I'll agree to keeping the introduction of `Util.bzero()` of the `prv_array` copy.

src/main/java/com/jcraft/jsch/jce/Util.java
norrisjeremy: This change is unneeded though to accomplish what you have proposed.
`src/main/java/com/jcraft/jsch/KeyPairECDSA.java` will be utilizing the class `src/main/java/com/jcraft/jsch/Util.java`, not `src/main/java/com/jcraft/jsch/jce/Util.java`.
Therefore there is no need for the changes to make `src/main/java/com/jcraft/jsch/jce/Util.java` public.

ikucuze:
> I'll agree to keeping the introduction of `Util.bzero()` of the `prv_array` copy.

Ok, pushed that. Copy of public key do not get cleaned.
Copy of private key is done, by making Util public.
pushed that now

Issue #752, opened 2025-01-23
Allow dumping JSch config for trouble shooting purposes


davsclaus:
The JSch class has a private config instance that is initialized with a lot of options.

However for trouble shooting then its hard to know and see the values of this config instance. There is a getConfig(key) api but then you need to provide every key name to get the value one by one.

Would it be possible to add an API that can return the known key names, so you can build some code that can get all the values for dumping / trouble shooting purposes.

The Apache Camel project would appreciated that.

norrisjeremy:
Hi @davsclaus,

You are welcome to submit a PR for us to review implementing such an API if it is important to you.

Thanks,
Jeremy

mwiede:
fixed with 0.2.23

davsclaus:
Thank you very much for the release - we have upgraded Camel and are working on using this new API for helping our users diagnose a bit better

PR #753, opened 2025-01-24
#752: Add getKeys to JSch which makes access to all the config values…

davsclaus:
… possible for troubleshooting.

### DIFF
--- a/src/main/java/com/jcraft/jsch/JSch.java
+++ b/src/main/java/com/jcraft/jsch/JSch.java
@@ -32,7 +32,6 @@ import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.Hashtable;
 import java.util.Map;
-import java.util.Set;
 import java.util.Vector;
 
 public class JSch {
--- a/src/main/java/com/jcraft/jsch/JSch.java
+++ b/src/main/java/com/jcraft/jsch/JSch.java
@@ -27,8 +27,10 @@
 package com.jcraft.jsch;
 
 import java.io.InputStream;
+import java.util.Collections;
 import java.util.Enumeration;
 import java.util.Hashtable;
+import java.util.Set;
 import java.util.Vector;
 
 public class JSch {
@@ -636,6 +638,20 @@ public class JSch {
       config.put("PubkeyAcceptedAlgorithms", value);
     } else {
       config.put(key, value);
+    }
+  }
+
+  /**
+   * Returns the config keys for all the stored configurations.
+   *
+   * For example this can be used for troubleshooting to enumerate all the configuration values to
+   * be logged.
+   *
+   * @return config keys
+   */
+  public static Set<String> getConfigKeys() {
+    synchronized (config) {
+      return Collections.unmodifiableSet(config.keySet());
     }
   }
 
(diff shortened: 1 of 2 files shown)


norrisjeremy (changes requested):
I think a new method like this would be more valuable, since it would provide users the entire config and not just the config keys:
```java
public static Map<String, String> getConfig() {
  Map<String, String> ret = new HashMap<>();
  synchronized (config) {
    for (Map.Entry<String, String> entry : config.entrySet()) {
      String key = entry.getKey();
      if (key.equals("PubkeyAcceptedKeyTypes")) {
        key = "PubkeyAcceptedAlgorithms";
      }
      ret.put(key, entry.getValue());
    }
  }
  return Collections.unmodifiableMap(ret);
}
```

davsclaus:
Thanks I have updated the PR

src/test/java/com/jcraft/jsch/JSchTest.java
```diff
@@ -64,6 +68,26 @@ void checkLoggerBehavior() throws Exception {
     assertSame(JSch.DEVNULL, jsch.getInstanceLogger(), "instance logger should be DEVNULL");
   }
 
+  @Test
+  void getConfigKeys() throws Exception {
```
norrisjeremy, line 72: Can we rename this test method, since it is now testing a method named `getConfig()` and not `getConfigKeys()`?

norrisjeremy (changes requested):

src/main/java/com/jcraft/jsch/JSch.java
```diff
@@ -27,8 +27,12 @@
 package com.jcraft.jsch;
 
 import java.io.InputStream;
+import java.util.Collections;
 import java.util.Enumeration;
+import java.util.HashMap;
 import java.util.Hashtable;
+import java.util.Map;
+import java.util.Set;
```
norrisjeremy, line 35: Can you remove this unused `java.util.Set` import that SonarQube is pointing out?

src/test/java/com/jcraft/jsch/JSchTest.java
```diff
@@ -64,6 +67,26 @@ void checkLoggerBehavior() throws Exception {
     assertSame(JSch.DEVNULL, jsch.getInstanceLogger(), "instance logger should be DEVNULL");
   }
 
+  @Test
+  void getConfig() throws Exception {
```
norrisjeremy, line 71: SonarQube points out that no method in this test case throws a checked Exception, so can we drop the `throws Exception` clause?

norrisjeremy (changes requested):

src/test/java/com/jcraft/jsch/JSchTest.java
davsclaus: all the other tests does the same

norrisjeremy (approved):
LGTM.

src/test/java/com/jcraft/jsch/JSchTest.java
norrisjeremy: Yes, I'll put on my TODO list to clean that up.

mwiede:
@norrisjeremy I will squash it when merging

davsclaus:
I think you can make GH have a button "squash and merge" that is what we do at Apache Camel

davsclaus:
created a new PR this can be closed

norrisjeremy (approved):
LGTM.

norrisjeremy:
> @norrisjeremy I will squash it when merging

Ok, that works too. Your choice whether to squash and merge this one or use the new #754.

davsclaus:
i suggest you try the "squash and merge" button

Issue #560, opened 2024-05-17
ConnectTimeout from ssh-config read as ms instead of seconds?


AlBundy33:
```
Host *
  ConnectTimeout 5
```
afaik the unit is seconds but it seems that JSch treats the value as milliseconds

mwiede:
Yes it seems to be an issue

https://github.com/mwiede/jsch/blob/7b7a2188d538569d4fec58824c4ef66a4405fd11/src/main/java/com/jcraft/jsch/OpenSSHConfig.java#L234-L237

norrisjeremy:
Hi @mwiede,

Do you think we should uncomment the code you cited?
From what I can see, it first appeared in the old 0.1.55 release from the original JSch author: I'm not sure what their reasoning was for keeping it commented out?

Thanks,
Jeremy

mwiede:
fixed with 0.2.23

PR #755, opened 2025-01-24
#560 treat openssh config values ConnectTimeout and ServerAliveInterval as seconds.

mwiede:
#560

### DIFF
--- a/src/main/java/com/jcraft/jsch/OpenSSHConfig.java
+++ b/src/main/java/com/jcraft/jsch/OpenSSHConfig.java
@@ -233,14 +233,15 @@ public class OpenSSHConfig implements ConfigRepository {
         if (value != null)
           break;
       }
-      // TODO: The following change should be applied,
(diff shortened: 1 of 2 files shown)


src/main/java/com/jcraft/jsch/OpenSSHConfig.java
```diff
@@ -303,6 +312,7 @@ public int getPort() {
         port = Integer.parseInt(foo);
       } catch (NumberFormatException e) {
         // wrong format
```
norrisjeremy, line 314: I think we can drop this `// wrong format` comment here: I believe I added it because some static analysis tool didn't like any empty catch statement, but that issue is now solved by the addition of `logError()` call.

mwiede: sure

norrisjeremy (approved):
LGTM.

PR #786, opened 2025-03-06
SSH server comparison the channel id from INT_MAX for a signed value.

asingh1508:
The problem is related to the checking of what is called a sender channel ID (which is set to a unsigned INT random number by the originator of the connection). However, in the versions mentioned above the daemon is making a comparison using INT_MAX for a signed value. This means that valid channel numbers are being rejected and the connection is not established.

### DIFF
--- a/src/main/java/com/jcraft/jsch/Channel.java
+++ b/src/main/java/com/jcraft/jsch/Channel.java
@@ -138,9 +138,9 @@ public abstract class Channel {
 
   Channel() {
     synchronized (pool) {
+      id = index++;
       // OpenSSH rejects channels with an id that exceeds INT_MAX
       index &= Integer.MAX_VALUE;
-      index++;
       pool.addElement(this);
     }
   }
--- a/src/main/java/com/jcraft/jsch/Channel.java
+++ b/src/main/java/com/jcraft/jsch/Channel.java
@@ -138,6 +138,7 @@ public abstract class Channel {
 
   Channel() {
     synchronized (pool) {
+      /* Restrict sender channel ID to be within INT_MAX */
       if (index == Integer.MAX_VALUE) {
         index = 0;
       }    
--- a/src/main/java/com/jcraft/jsch/Channel.java
+++ b/src/main/java/com/jcraft/jsch/Channel.java
@@ -139,7 +139,7 @@ public abstract class Channel {
   Channel() {
     synchronized (pool) {
       id = index;
-      index = (index == Integer.MAX_VALUE) ? 0 : ++index; 
+      index = index == Integer.MAX_VALUE ? 0 : ++index; 
       pool.addElement(this);
     }
   }
--- a/src/main/java/com/jcraft/jsch/Channel.java
+++ b/src/main/java/com/jcraft/jsch/Channel.java
@@ -140,6 +140,7 @@ public abstract class Channel {
     synchronized (pool) {
       // OpenSSH rejects channels with an id that exceeds INT_MAX
       index &= Integer.MAX_VALUE;
+      index++;
       pool.addElement(this);
     }
   }
--- a/src/main/java/com/jcraft/jsch/Channel.java
+++ b/src/main/java/com/jcraft/jsch/Channel.java
@@ -138,7 +138,7 @@ public abstract class Channel {
 
   Channel() {
     synchronized (pool) {
-      /* Restrict sender channel ID to be within INT_MAX */
+      /* Restrict sender channel ID to be within INT_MAX value */
       if (index == Integer.MAX_VALUE) {
         index = 0;
       }    
--- a/src/main/java/com/jcraft/jsch/Channel.java
+++ b/src/main/java/com/jcraft/jsch/Channel.java
@@ -143,7 +143,7 @@ public abstract class Channel {
         index = 0;
       } else {
         ++index;
-      }  
+      }
       pool.addElement(this);
     }
   }
--- a/src/main/java/com/jcraft/jsch/Channel.java
+++ b/src/main/java/com/jcraft/jsch/Channel.java
@@ -138,11 +138,11 @@ public abstract class Channel {
 
   Channel() {
     synchronized (pool) {
-      if (index == Integer.MAX_VALUE) {
+      if (index == 2147483647) {
         index = 0;
       }
       id = index++;
-      pool.addElement(this);      
+      pool.addElement(this);
     }
   }
 
--- a/src/main/java/com/jcraft/jsch/Channel.java
+++ b/src/main/java/com/jcraft/jsch/Channel.java
@@ -139,13 +139,11 @@ public abstract class Channel {
   Channel() {
     synchronized (pool) {
       id = index++;
-      
       // OpenSSH 8.0 introduced a bug that rejected channels with an ID that exceeds INT_MAX.
       // See https://github.com/openssh/openssh-portable/commit/7ec5cb4.
       // This bug was later resolved in OpenSSH 8.2.
       // See https://github.com/openssh/openssh-portable/commit/0ecd20b.
       // To allow compability, cap the ID value to not exceed INT_MAX.
-      
       index &= Integer.MAX_VALUE;
       pool.addElement(this);
     }
--- a/src/main/java/com/jcraft/jsch/Channel.java
+++ b/src/main/java/com/jcraft/jsch/Channel.java
@@ -138,10 +138,10 @@ public abstract class Channel {
 
   Channel() {
     synchronized (pool) {
+      id = index++;
       if (index == Integer.MAX_VALUE) {
         index = 0;
       }
-      id = index++;
       pool.addElement(this);
     }
   }
--- a/src/main/java/com/jcraft/jsch/Channel.java
+++ b/src/main/java/com/jcraft/jsch/Channel.java
@@ -138,6 +138,9 @@ public abstract class Channel {
 
   Channel() {
     synchronized (pool) {
+      if (index == Integer.MAX_VALUE) {
+        index = 0;
+      }    
       id = index++;
       pool.addElement(this);
     }
--- a/src/main/java/com/jcraft/jsch/Channel.java
+++ b/src/main/java/com/jcraft/jsch/Channel.java
@@ -139,7 +139,11 @@ public abstract class Channel {
   Channel() {
     synchronized (pool) {
       id = index;
-      index = index == Integer.MAX_VALUE ? 0 : ++index; 
+      if (index == Integer.MAX_VALUE) {
+        index = 0;
+      } else {
+        ++index;
+      }  
       pool.addElement(this);
     }
   }
--- a/src/main/java/com/jcraft/jsch/Channel.java
+++ b/src/main/java/com/jcraft/jsch/Channel.java
@@ -139,7 +139,8 @@ public abstract class Channel {
   Channel() {
     synchronized (pool) {
       id = index++;
-      // OpenSSH rejects channels with an id that exceeds INT_MAX
+      // OpenSSH versions 8.0 and 8.1 reject channels with an ID that exceeds INT_MAX.
+      // It had resolved it in OpenSSH versions after to 8.1(https://github.com/openssh/openssh-portable/commit/0ecd20bc9f0b9c7c697c9eb014613516c8f65834).
       index &= Integer.MAX_VALUE;
       pool.addElement(this);
     }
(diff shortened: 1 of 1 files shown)


norrisjeremy:
Hi @asingh1508,

Can you provide more details as to what problem this is solving?
Is this an issue with some specific SSH server implementation that fails with channel id values that exceed INT_MAX?

Thanks,
Jeremy

asingh1508:
@norrisjeremy 

We're describing a bug in OpenSSH where the OpenSSH daemon (sshd) incorrectly handles sender channel IDs due to an issue with signed vs. unsigned integer comparisons.

**Understanding the Issue:**

In SSH, each side of a connection assigns a sender channel ID, which is an unsigned integer (typically a random number).
OpenSSH daemon (sshd) checks this channel ID when establishing a connection.

However, in the affected versions, sshd mistakenly compares the sender channel ID using INT_MAX (which is for signed integers).
This means that if the sender channel ID is greater than INT_MAX (i.e., exceeds 2147483647 on a 32-bit system), the check fails.
As a result, valid connections are being rejected.

**Impact:**
This bug can prevent SSH connections from being established in certain cases.
It might affect systems with high activity or specific configurations where channel IDs can exceed the signed integer limit.

OpenSSH server code:
/* Parse a channel ID from the current packet */
static int
channel_parse_id(struct ssh *ssh, const char *where, const char *what)
{
        **u_int32_t id;**
        int r;



    if ((r = sshpkt_get_u32(ssh, &id)) != 0) {
            error("%s: parse id: %s", where, ssh_err(r));
            ssh_packet_disconnect(ssh, "Invalid %s message", what);
    }
    **if (id > INT_MAX) {**
            error("%s: bad channel id %u: %s", where, id, ssh_err(r));
            ssh_packet_disconnect(ssh, "Invalid %s channel id", what);
    }
    return (int)id;
}

asingh1508:
> Hi @asingh1508,
> 
> Can you provide more details as to what problem this is solving? Is this an issue with some specific SSH server implementation that fails with channel id values that exceed INT_MAX?
> 
> Thanks, Jeremy




> Is this an issue with some specific SSH server implementation that fails with channel id values that exceed INT_MAX?
> 
> Thanks,

Its not specific server SSH server. This issue will happend all  SSH to servers that are running OpenSSH daemon.

norrisjeremy (changes requested):
Hi @asingh1508,

Thank you for providing information on how OpenSSH rejects channels with an id that exceeds INT_MAX.

I think a cleaner fix would be something like this: can you confirm and if so, adjust this PR?
```diff
diff --git a/src/main/java/com/jcraft/jsch/Channel.java b/src/main/java/com/jcraft/jsch/Channel.java
index 6fb79a7..467d591 100644
--- a/src/main/java/com/jcraft/jsch/Channel.java
+++ b/src/main/java/com/jcraft/jsch/Channel.java
@@ -139,6 +139,8 @@ public abstract class Channel {
   Channel() {
     synchronized (pool) {
       id = index++;
+      // OpenSSH rejects channels with an id that exceeds INT_MAX
+      index &= Integer.MAX_VALUE;
       pool.addElement(this);
     }
   }
```

Thanks,
Jeremy

src/main/java/com/jcraft/jsch/Channel.java
```diff
@@ -138,7 +138,8 @@ static void del(Channel c) {
 
   Channel() {
     synchronized (pool) {
-      id = index++;
```
norrisjeremy, line 141: This `id = index++;` line needs to be restored.

norrisjeremy (changes requested):

src/main/java/com/jcraft/jsch/Channel.java
```diff
@@ -138,7 +138,9 @@ static void del(Channel c) {
...
+      // OpenSSH rejects channels with an id that exceeds INT_MAX
+      index &= Integer.MAX_VALUE;
+      index++;
```
norrisjeremy, line 143: This is still incorrect.
This all needs to be like this:
```java
    synchronized (pool) {
      id = index++;
      // OpenSSH rejects channels with an id that exceeds INT_MAX
      index &= Integer.MAX_VALUE;
      pool.addElement(this);
    }
```

norrisjeremy (changes requested):

norrisjeremy (approved):
LGTM, thanks!

asingh1508:
> LGTM, thanks!

@norrisjeremy Who will this pull request merged in master branch ?

norrisjeremy:
> > LGTM, thanks!
> 
> @norrisjeremy Who will this pull request merged in master branch ?

We will need to wait for @mwiede.

norrisjeremy (changes requested):
Actually, I am reviewing OpenSSH code more closely and more changes may be required.
I will respond back once I have completed a more thorough review.

asingh1508:
> Actually, I am reviewing OpenSSH code more closely and more changes may be required. I will respond back once I have completed a more thorough review.

sure

norrisjeremy:
Hi @asingh1508,

Actually, after reviewing further, I'm not sure if I believe any changes are actually needed.
I believe the code you link is related to handling of the channel id value that the OpenSSH server assigns to the channel, and not the id that the JSch client assigns to the channel.

All SSH channels consist of two channel id values: one assigned by the client and one assigned by the server (see [RFC-4254 Section 5.1](https://datatracker.ietf.org/doc/html/rfc4254#section-5.1)).

The code you cited in OpenSSH appears to be related to handling of it's locally assigned id value that it associates with the channel and not the id value that JSch associates with the channel.

Have you actually experienced a failure that can somehow be traced back to JSch channel id values exceeding `0x7fffffff` (`INT_MAX`), or did you submit this based purely on code inspection?

Thanks,
Jeremy

asingh1508:
> Have you actually experienced a failure that can somehow be traced back to JSch channel id values exceeding `0x7fffffff` (`INT_MAX`),

I had experienced a failure in customer machine where JSch channel id values exceeding `0x7fffffff` (`INT_MAX`), 

> you submit this based purely on code inspection

@norrisjeremy 
I had experienced ssh connection failure in customer machine where JSch channel id values exceeding 0x7fffffff INT_MAX),.
This was reason i inspectioned the code and summit this pr.

norrisjeremy:
> > Have you actually experienced a failure that can somehow be traced back to JSch channel id values exceeding `0x7fffffff` (`INT_MAX`),
> 
> I had experienced a failure in customer machine where JSch channel id values exceeding `0x7fffffff` (`INT_MAX`),
> 
> > you submit this based purely on code inspection
> 
> @norrisjeremy I had experienced ssh connection failure in customer machine where JSch channel id values exceeding 0x7fffffff INT_MAX),. This was reason i inspectioned the code and summit this pr.

Can you share more details as to how you arrived at the conclusion that the failure you experienced is related to a JSch channel id exceeding 0x7fffffff (INT_MAX)?
Because I'm not convinced by my further examination of the OpenSSH code that it should result in a failure.

norrisjeremy:
Hi @asingh1508,

But I do not believe the code you cited is actually the code that is used to parse the `SSH2_MSG_CHANNEL_OPEN` that JSch sends.
I urge you to more carefully read and review the OpenSSH code, in particular the code in the `server_input_channel_open()` function present in `serverloop.c`, that is responsible for parsing and handling the remote channel id from the sender (stored in the `rchan` variable).
Additionally, I have reviewed a couple other SSH implementations and none them appear to clamp the id value they use in the `SSH2_MSG_CHANNEL_OPEN` to ensure that it doesn't exceed `INT_MAX`, so I have a difficult time believing this is actually an issue.

Thanks,
Jeremy

asingh1508:
> server_input_channel_open

static int
server_input_channel_open(int type, u_int32_t seq, struct ssh *ssh)
{
        Channel *c = NULL;
        char *ctype = NULL;
        const char *errmsg = NULL;
        int r, reason = SSH2_OPEN_CONNECT_FAILED;
        u_int rchan = 0, rmaxpack = 0, rwindow = 0;

        if ((r = sshpkt_get_cstring(ssh, &ctype, NULL)) != 0 ||
            (r = sshpkt_get_u32(ssh, &rchan)) != 0 ||
            (r = sshpkt_get_u32(ssh, &rwindow)) != 0 ||
            (r = sshpkt_get_u32(ssh, &rmaxpack)) != 0)
                sshpkt_fatal(ssh, r, "%s: parse packet", __func__);
        debug("%s: ctype %s rchan %u win %u max %u", __func__,
            ctype, rchan, rwindow, rmaxpack);

        **if (rchan > INT_MAX) {**
                error("%s: invalid remote channel ID", __func__);

@norrisjeremy rchan is  u_int type variable but it is comparing with if (rchan > INT_MAX). 
@norrisjeremy **Can you please check for openssh-8.1p1 and openssh-8.0p1 version ?**

asingh1508:
> openssh-8.1p1 and openssh-8.0p1 version

norrisjeremy (approved):
Hi @asingh1508,

Ok, I now see that this was a bug that existed until OpenSSH 8.2, when it was fixed in https://github.com/openssh/openssh-portable/commit/0ecd20bc9f0b9c7c697c9eb014613516c8f65834.

I would encourage you in the future when opening issues or PRs with us, to include as many details as possible, to help us understand the conditions in which you were encountering as issue.

For example, pinpointing the fact that this impacted older OpenSSH releases was crucial in me understanding why this workaround is needed.

Thanks,
Jeremy

src/main/java/com/jcraft/jsch/Channel.java
```diff
@@ -139,6 +139,8 @@ static void del(Channel c) {
   Channel() {
     synchronized (pool) {
       id = index++;
+      // OpenSSH rejects channels with an id that exceeds INT_MAX
```
norrisjeremy, line 142: We should update this comment to reflect that OpenSSH versions prior to 8.2 were affected by this issue.
And possible include a link to the changeset id that resolved it (https://github.com/openssh/openssh-portable/commit/0ecd20bc9f0b9c7c697c9eb014613516c8f65834).

norrisjeremy (changes requested):

src/main/java/com/jcraft/jsch/Channel.java
asingh1508: @norrisjeremy done. Please verify

src/main/java/com/jcraft/jsch/Channel.java
```diff
@@ -139,6 +139,9 @@ static void del(Channel c) {
   Channel() {
     synchronized (pool) {
       id = index++;
+      // OpenSSH versions 8.0 and 8.1 reject channels with an ID that exceeds INT_MAX.
+      // It had resolved it in OpenSSH versions after to 8.1(https://github.com/openssh/openssh-portable/commit/0ecd20bc9f0b9c7c697c9eb014613516c8f65834).
+      index &= Integer.MAX_VALUE;
```
norrisjeremy, line 144: Let's improve the wording for this:
```java
      /*
       * OpenSSH 8.0 introduced a bug that rejected channels with an ID that exceeds INT_MAX.
       * See https://github.com/openssh/openssh-portable/commit/7ec5cb4d15ed2f2c5c9f5d00e6b361d136fc1e2d.
       * This bug was later resolved in OpenSSH 8.2.
       * See https://github.com/openssh/openssh-portable/commit/0ecd20bc9f0b9c7c697c9eb014613516c8f65834.
       * To ensure compatibility, clamp the ID value JSch uses to not exceed INT_MAX.
       */
```

norrisjeremy (approved):

src/main/java/com/jcraft/jsch/Channel.java
norrisjeremy: Actually, let's format it like this to appease the `formatter-maven-plugin`:
```java
      // OpenSSH 8.0 introduced a bug that rejected channels with an ID that exceeds INT_MAX.
      // See https://github.com/openssh/openssh-portable/commit/7ec5cb4.
      // This bug was later resolved in OpenSSH 8.2.
      // See https://github.com/openssh/openssh-portable/commit/0ecd20b.
      // To allow compability, cap the ID value to not exceed INT_MAX.
```

norrisjeremy (changes requested):

src/main/java/com/jcraft/jsch/Channel.java
asingh1508: > Actually, let's format it like this to appease the `formatter-maven-plugin`:
> 
> ```java
>       // OpenSSH 8.0 introduced a bug that rejected channels with an ID that exceeds INT_MAX.
>       // See https://github.com/openssh/openssh-portable/commit/7ec5cb4.
>       // This bug was later resolved in OpenSSH 8.2.
>       // See https://github.com/openssh/openssh-portable/commit/0ecd20b.
>       // To allow compability, cap the ID value to not exceed INT_MAX.
> ```

Done

asingh1508:
@norrisjeremy Buid with all tests passed succesfully. Please verify and approved..

norrisjeremy (approved):
Thanks for you patience.
I think this looks good now.

PR #793, opened 2025-03-16
Support custom ThreadFactory.

hstyi:
This PR adds support for a custom `ThreadFactory` in JSch. Users can now provide their own `ThreadFactory`, which is particularly useful in Java 21, as it allows leveraging virtual threads for better concurrency and resource management.

### DIFF
--- a/src/main/java/com/jcraft/jsch/Session.java
+++ b/src/main/java/com/jcraft/jsch/Session.java
@@ -40,6 +40,8 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Properties;
 import java.util.Vector;
+import java.util.concurrent.ThreadFactory;
+import java.util.Objects;
 import javax.crypto.AEADBadTagException;
 
 public class Session {
@@ -173,6 +175,8 @@ public class Session {
 
   JSch jsch;
   Logger logger;
+
+  private ThreadFactory threadFactory = Thread::new;
 
   Session(JSch jsch, String username, String host, int port) throws JSchException {
     super();
@@ -528,7 +532,7 @@ public class Session {
 
       synchronized (lock) {
         if (isConnected) {
-          connectThread = jsch.getThreadFactory().newThread(this::run);
+          connectThread = getThreadFactory().newThread(this::run);
           connectThread.setName("Connect thread " + host + " session");
           if (daemon_thread) {
             connectThread.setDaemon(daemon_thread);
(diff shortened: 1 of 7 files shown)


norrisjeremy (changes requested):
So a few suggestions:

1. Instead of using `Executors.defaultThreadFactory()`, let's instead just use a simple lambda based `Thread::new` as the default, as this will functionally make this a noop change (since `Executors.defaultThreadFactory()` sets a Thread name, etc.).
2. Let's initialize the new `threadFactory` property to a non-null value and then disallow users from ever setting it to null via `setThreadFactory()`, so that we never have to explicitly check for null in `getThreadFactory()`.

I would also caution against using JSch with virtual threads, at least unless you are using the new Java 24 release that came out earlier this week, since I suspect you would likely encounter problems.
JSch makes significant use of `synchronized` style locking, that until the new Java 24 release results in carrier thread pinning when combined with virtual threads, which in turn can lead to difficult to diagnose deadlocks.

hstyi:
@norrisjeremy Thank you for your reply.

1. We can change `synchronized` to `ReentrantLock`.
2. Putting `threadFactory` in the Session class does allow for more granular control.

In addition, JDK 21 is being used more and more, especially in high IO usage scenarios where virtual threads are very advantageous.

hstyi:
Hi @norrisjeremy ,

I've moved the `ThreadFactory` to `Session`.

As far as I know, Java24([JEP491](https://openjdk.org/jeps/491)) has been released. It has solved the problems that could occur with virtual threads using `synchronized`.

src/main/java/com/jcraft/jsch/Session.java
```diff
@@ -2156,6 +2160,27 @@ public void disconnect() {
...
+    this.threadFactory = Objects.requireNonNull(threadFactory);
+  }
+
+
+  /**
+   * Returns the thread factory used by this instance.
+   *
+   * @return The thread factory associated with this instance. If no specific thread factory has
+   *         been set, a default thread factory is created and returned.
+   */
```
norrisjeremy, line 2179:
```suggestion
  /**
   * Returns the thread factory used by this instance.
   *
   * @return The thread factory associated with this instance. If no specific thread factory has
   *         been set, a default thread factory is returned.
   */
```

src/main/java/com/jcraft/jsch/Session.java
```diff
@@ -40,6 +40,8 @@
 import java.util.Locale;
 import java.util.Properties;
 import java.util.Vector;
+import java.util.concurrent.ThreadFactory;
+import java.util.Objects;
```
norrisjeremy, line 44:
```suggestion
import java.util.Locale;
import java.util.Objects;
import java.util.Properties;
import java.util.Vector;
import java.util.concurrent.ThreadFactory;
```

norrisjeremy (approved):

norrisjeremy:
@mwiede Can you retrigger a build of this PR to confirm the SonarCloud check passes with your changes?

PR #797, opened 2025-03-20
0.2.25 changes

norrisjeremy:
- Add support for mlkem768x25519-sha256, mlkem768nistp256-sha256 & mlkem1024nistp384-sha384 KEX algorithms using JEP 496.
- Stop abusing the packet buffer for signature verification.
- Fix intermittent KEX errors with hybrid PQ/EC algorithms: the EC shared secret should not be stripped of unnecessary leading zero bytes since they are concatenated with the PQ secret and encoded as a string.
- Update forbiddenapis to use latest commons-io signatures.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -256,7 +256,7 @@
                         <configuration>
                             <rules>
                                 <requireJavaVersion>
-                                    <version>17</version>
+                                    <version>24</version>
                                 </requireJavaVersion>
                                 <dependencyConvergence />
                                 <banDuplicatePomDependencyVersions />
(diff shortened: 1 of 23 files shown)


norrisjeremy:
FYI, I believe this is finally ready to go.
I was able to finally track down the source of the intermittent failures that were occurring with the PQ/EC algorithms.

norrisjeremy:
> 21.1% Duplication on New Code

@mwiede Can we disable this SonarQube check?
I feel it's `Duplication on New Code` check isn't especially helpful: I believe that trying to massively deduplicate code in this library to appease it would carry far too great of a risk for regressions.

pom.xml
```diff
@@ -836,15 +851,16 @@
...
             </activation>
             <build>
                 <plugins>
                     <plugin>
                         <groupId>de.thetaphi</groupId>
                         <artifactId>forbiddenapis</artifactId>
                         <configuration>
-                            <releaseVersion>16</releaseVersion>
+                            <!-- This should be bumped to 24 when forbiddenapis catches up -->
+                            <releaseVersion>23</releaseVersion>
```
norrisjeremy, line 863: If we can get #806 merged, I can rebase on top of that and then fix this.

mwiede:
@norrisjeremy I adopted the sonar quality gate, which allows 25% of duplicated code. Lets see, whether it becomes green with next analysis

norrisjeremy:
> @norrisjeremy I adopted the sonar quality gate, which allows 25% of duplicated code. Lets see, whether it becomes green with next analysis

Great, thanks!
I think this is ready to land now.

Issue #818, opened 2025-04-15
0.2.25 was bulit with java 24 as target level


turing85:
Hello!
We tried to upgrade `jsch` from `0.2.24` to `0.2.25` and got the following:

```
org.springframework.core.NestedIOException: ASM ClassReader failed to parse class file - probably due to a new Java class file version that isn't supported yet: URL [jar:file:/opt/osiris/spring-boot/app.jar!/META-INF/versions/24/com/jcraft/jsch/jce/MLKEM.class]; nested exception is java.lang.IllegalArgumentException: Unsupported class file major version 68
```

I see that the java-version was changed to `24` in https://github.com/mwiede/jsch/commit/303785be7777864f56f1e5ea3ce01f4dcae2286c. It looks to me as if the target level should be configured to still be `21` (or `17`?).

it would be nice to have releases that use an LTS-version as target level.

norrisjeremy:
Hi @turing85,

JSch uses a multi-release jar-file. Some parts of it are compiled specifically targeting Java 24 in order to support some new features, but the base part of the library still targets Java 8 and is still fully usable with Java 8 (we fully test this with all of our unit & integration tests run in Github Actions CI).

I'm not sure why your Spring app is attempting to use ASM to read the Java 24 part of the multi-release jar-file, but I would recommend taking it up with the Spring development team.
Or you could simply update the version of ASM you are using, as the latest release fully supports Java 24 class files.

Thanks,
Jeremy

pschulz:
I think there is some definition missing in https://github.com/mwiede/jsch/blob/master/src/assembly/sources.xml in order to keep the compatibility, so that these v24 specific classes become put into META-INF/versions/24.

I have the same issue with 0.2.26, but in another context. The result is the same and the runtime thinks it needs Java 24.

norrisjeremy:
Hi @pschulz,

I have hopefully correct the oversight you have noted as part of #837 (see 86fce97).

Thanks,
Jeremy

norrisjeremy:
Hi @pschulz,

I'm not sure what is leading you to believe that the Java 24 classes are not located in `META-INF/versions/24`?

```shell
$ wget https://repo1.maven.org/maven2/com/github/mwiede/jsch/2.27.0/jsch-2.27.0.jar
--2025-06-06 15:55:18--  https://repo1.maven.org/maven2/com/github/mwiede/jsch/2.27.0/jsch-2.27.0.jar
[...]

PR #838, opened 2025-06-06
Update assembly/sources.xml

pschulz:
Bug fix for #818

Ensure we have a correct multi-release JAR.


norrisjeremy:
This is already handled as part of #837, so it can be closed.

PR #887, opened 2025-08-21
Slowness due to blocked threads in Channel.getChannel call

DavidTavoularis:


norrisjeremy (changes requested):
The strategy that is being proposed here could potentially result in memory leaks due to stranded Channels.

For example, the old strategy, while likely slow as you have pointed out, there was no chance for a Channel to not be added to the pool, even if they happened to somehow share the same id value.

However, the strategy proposed here, if two Channel's ever ended up sharing the same id value, the later created one would overwrite it's position in the pool & thus result in a memory leak (since the earlier created Channel would no longer be present in the pool).

src/main/java/com/jcraft/jsch/Channel.java
```diff
@@ -137,16 +133,19 @@ static void del(Channel c) {
...
+    // See https://github.com/openssh/openssh-portable/commit/0ecd20b.
+    // To allow compability, cap the ID value to not exceed INT_MAX.
+    int currentIndex;
+    int nextIndex;
+    do {
+      currentIndex = index.get();
+      nextIndex = (currentIndex + 1) & Integer.MAX_VALUE;
+    } while (!index.compareAndSet(currentIndex, nextIndex));
+    id = currentIndex;
+    pool.put(id, this);
```
norrisjeremy: To illustrate, what I mean, suppose at time X, the `id` value for a particular Channel A was computed as a value of 5.
Then suppose at some later time Y, the `id` value for a different Channel B happens to be computed as a value of 5 (since the AtomicInteger `index` value cycled through enough values to rollover and begin to be recycled).
The `pool.put(id, this)` call here that puts Channel B into into the pool with an `id` key value of 5 overwrites the previous entry with a key value of 5 that points at the earlier Channel A.
And thus we have a memory leak, since Channel A is no longer being maintained in the `pool`.

norrisjeremy:
Hi @DavidTavoularis,

I'm not a fan of either solution honestly, because it artificially restricts the reuse of a specific id values.
I'm working on an alternate solution to see if it would be possible to remedy the issue you have reported by simply using a [ReadWriteLock](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/ReadWriteLock.html).
I'll report back once I've had some time to perform more testing.

Thanks,
Jeremy

DavidTavoularis:
Hi @norrisjeremy thank you for suggesting to use ReadWriteLock.

I just updated the pull request with these new changes:
  - ReadWriteLock for concurrent access control
  - ArrayList instead of Vector (no need for Vector's synchronization)
  - Read lock for lookups - allows multiple concurrent reads
  - Write lock for modifications - exclusive access when adding/removing

  Benefits:
  1. Better concurrency - Multiple threads can search channels simultaneously
  2. No ID collision issues - IDs can be freely reused as channels are removed
  3. No memory leaks - Every channel is tracked in the list
  4. Cleaner code - Using modern ArrayList with explicit locking instead of legacy Vector
  5. Maintains original behavior - ID overflow handling preserved exactly

  Key operations:
  - getChannel() - Uses read lock, allows concurrent lookups
  - Channel() constructor - Uses write lock for adding to pool
  - del() - Uses write lock for removal
  - disconnect() - Uses read lock to collect channels, then disconnects outside the lock

This solution provides thread safety without the artificial restrictions of collision detection, while improving performance for concurrent read operations.

norrisjeremy:
> Hi @DavidTavoularis,
> 
> See #888 for my current draft proposal.
> 
> Thanks, Jeremy

FYI, I still need to think through the memory usage ramifications of changing the `pool` structure from a [Vector](https://docs.oracle.com/javase/8/docs/api/java/util/Vector.html) to either an [ArrayList](https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html) as you have proposed or a [HashMap](https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html) as I have done in my draft PR.

DavidTavoularis:
@norrisjeremy I like your approach, here are my comments, regarding your pull-request:

1. Redundant ID check in getChannel:
if (c.id == id && c.session == session)
Since you're already looking in pool.get(id), the c.id == id check is redundant. Should be:
if (c.session == session)

2. SonarQube: Unused imports - You have Collection imported but not used
The other SonarQube "Remove this assignment of "index" should be flagged as "OK" (as no change compared to before)

3. Should you put pool & poolLock as final ?

4. While this new implementation handles collisions, the Set<Channel> approach adds small memory overhead of maintaining Sets even for single channels, so should you replace:
pool.computeIfAbsent(id, k -> new HashSet<>()).add(this);
by:
pool.computeIfAbsent(id, k -> new HashSet<>(1)).add(this);
as 99.99999% of the time, the HashSet size will be 1 ?

Best Regards

src/main/java/com/jcraft/jsch/Channel.java
```diff
@@ -44,8 +48,11 @@ public abstract class Channel {
   static final int SSH_OPEN_UNKNOWN_CHANNEL_TYPE = 3;
   static final int SSH_OPEN_RESOURCE_SHORTAGE = 4;
 
-  static int index = 0;
-  private static Vector<Channel> pool = new Vector<>();
+  private static final ReadWriteLock poolLock = new ReentrantReadWriteLock();
+  private static final Lock readLock = poolLock.readLock();
+  private static final Lock writeLock = poolLock.writeLock();
```
norrisjeremy: I would omit these two properties and simply derive the read & write locks in the methods that need to use them like I do in #888.

DavidTavoularis:
@norrisjeremy Local ephemeral port range (typically ~28,000-65,000 ports)
Is it true that each Channel is using a different local ephemeral port range ?

If yes, then the maximum amount of simultaneous Channels is about 50000, but I don't believe that anyone would use more than 10000 simultaneous Channels, a more typical usage would be 1000 simultaneous Channels for high-performance systems (this is my use-case).
Using Map<Integer, Set> would increase memory usage by 800KB for 10000 Channels and by only 80KB for 1000 Channels.

norrisjeremy:
> @norrisjeremy Local ephemeral port range (typically ~28,000-65,000 ports) Is it true that each Channel is using a different local ephemeral port range ?
> 
> If yes, then the maximum amount of simultaneous Channels is about 50000, but I don't believe that anyone would use more than 10000 simultaneous Channels, a more typical usage would be 1000 simultaneous Channels for high-performance systems (this is my use-case). Using Map<Integer, Set> would increase memory usage by 800KB for 10000 Channels and by only 80KB for 1000 Channels.

The id values are only locally significant to a particular session.
You could have thousands of sessions, and in each session have thousands of channels.

DavidTavoularis:
I believe that there is usually a limit of 10 channels per SSH session (cf MaxSessions 10 in /etc/ssh/sshd_config)

src/main/java/com/jcraft/jsch/Channel.java
```diff
@@ -609,22 +625,23 @@ public boolean isClosed() {
...
+    l.lock();
+    try {
+      for (Channel c : pool) {
         try {
-          Channel c = pool.elementAt(i);
           if (c.session == session) {
-            channels[count++] = c;
+            channels.add(c);
           }
         } catch (Exception e) {
```
norrisjeremy: I'm not sure why this inner try block that catches exceptions & ignores them exists.
I wonder if we should just drop it?

DavidTavoularis: agreed, I dropped it and pushed in PR

src/main/java/com/jcraft/jsch/Channel.java
```diff
@@ -44,8 +48,9 @@ public abstract class Channel {
   static final int SSH_OPEN_UNKNOWN_CHANNEL_TYPE = 3;
   static final int SSH_OPEN_RESOURCE_SHORTAGE = 4;
 
-  static int index = 0;
-  private static Vector<Channel> pool = new Vector<>();
+  private static final ReadWriteLock poolLock = new ReentrantReadWriteLock();
+  private static int index = 0;
+  private static List<Channel> pool = new ArrayList<>();
```
norrisjeremy: I think pool could be marked final?

DavidTavoularis:
> I think going with this more conservative approach is safer than the alternate approach I explored in #888, so I think we should go with this.

agreed

norrisjeremy:
@mwiede I think what we should do is merge this PR (#887), and then I will rebase my PR (#889) on top of the changes in this, and then I will take #889 out of draft status and you can then merge it.
That will give us the ability to easily rollback the changes in #889 in case we discover any sort of regression.

Issue #885, opened 2025-08-21
Slowness due to blocked threads in Channel.getChannel call in multi-thread environment


DavidTavoularis:
On a quite old jsch release (0.1.55), we could see lots of blocked threads in multi-thread environment.
This is due to the fact that Channel is using a Vector<Channel> for its internal pool.

The issue is still there on master, so I have prepared a merge-request, pending your review and approval.

Note1:  The code change was tested fine in lab and production (on top of 0.1.55).
The suggested code change had to manage this new line (not present in 0.1.55):
index &= Integer.MAX_VALUE;
which was introduced recently with https://github.com/mwiede/jsch/commit/793753a9897421f107e9c8dadd13410c5e481cd4
So it required some specific change to be thread-safe (and not tested in production).

Note2: After code change on top of 0.1.55, it was observed no more slowness or blocked threads

Here are the same observed stack-traces (every minute) showing the blocked thread when calling Channel.getChannel:
Stack 4 - Mar 05 05:54:01
 "Connect thread X.X.X.X session" #2631 prio=5 os_prio=64 tid=0x0000000106346800 nid=0xa51 waiting for monitor entry [0xfffffffe31bfe000]
    java.lang.Thread.State: BLOCKED (on object monitor)
                 at com.jcraft.jsch.Channel.getChannel(Channel.java:83)
                 - waiting to lock <0x00000004e52a0168> (a java.util.Vector)
                 at com.jcraft.jsch.Session.run(Session.java:1448)
                 at java.lang.Thread.run(Thread.java:745)
  
 Stack 5 - Mar 05 05:55:00
 "Connect thread X.X.X.X session" #2631 prio=5 os_prio=64 tid=0x0000000106346800 nid=0xa51 waiting for monitor entry [0xfffffffe31bfe000]
    java.lang.Thread.State: BLOCKED (on object monitor)
                 at com.jcraft.jsch.Channel.getChannel(Channel.java:83)
                 - waiting to lock <0x00000004e52a0168> (a java.util.Vector)
                 at com.jcraft.jsch.Session.run(Session.java:1448)
                 at java.lang.Thread.run(Thread.java:745)
  
 Stack 6 - Mar 05 05:56:00
 "Connect thread X.X.X.X session" #2631 prio=5 os_prio=64 tid=0x0000000106346800 nid=0xa51 waiting for monitor entry [0xfffffffe31bfe000]
    java.lang.Thread.State: BLOCKED (on object monitor)
[...]

PR #889, opened 2025-08-25
Store Channel objects assocaited with a particular Session with the S…

norrisjeremy:
…ession instance in order to try and improve performance of Channel lookups by id. (#885)


src/main/java/com/jcraft/jsch/Session.java
```diff
@@ -2101,7 +2137,17 @@ public void disconnect() {
...
      */
 
-    Channel.disconnect(this);
+    Lock l = channelsLock.writeLock();
+    l.lock();
+    try {
+      Iterator<Channel> it = channels.iterator();
+      while (it.hasNext()) {
+        it.next()._disconnect();
+        it.remove();
```
norrisjeremy: This part is my biggest concern, since we would be performing IO with the lock held.
I don't believe there's any chance for deadlocks from what I can see, but it still leaves me a bit uncomfortable.

DavidTavoularis:
@norrisjeremy Indeed, I prefer this new approach
 
1. Regarding getChannelById() method (Session.java:1142-1155), should you replace "c.id == id && c.session == this" by "c.id == id" as session check is redundant since channels are now stored per-session
 
2. Regarding delChannel() method (Session.java:2111-2118), should you just do "channels.remove(c)" instead of using an iterator ?

3. Regarding disconnect() method (Session.java:2137-2146), should you avoid iterator ?
from :
```
Iterator<Channel> it = channels.iterator();
      while (it.hasNext()) {
        it.next()._disconnect();
        it.remove();
      }
```
to :
```
      for (Channel channel : channels) {
        channel._disconnect();
      }
      channels.clear();
```

4. Why did you add "_" in _disconnect method name (cf SonarQloud) ?

5. Regarding your concern, one solution would be to add "volatile boolean disconnecting = false;" in Channel.java and "if (c.id == id && !c.disconnecting)" in Session.getChannelById method and perform this change in Session.disconnect() method:
```
    // Mark all channels as disconnecting
    List<Channel> channelsToDisconnect;
    Lock l = channelsLock.writeLock();
    l.lock();
    try {
      channelsToDisconnect = new ArrayList<>(channels);
      for (Channel channel : channels) {
        channel.disconnecting = true;
      }
    } finally {
      l.unlock();
    }

    // Disconnect channels without holding the lock (I/O operations)
    for (Channel channel : channelsToDisconnect) {
      channel._disconnect();
    }
    
    // Remove all channels after disconnection is complete
    l = channelsLock.writeLock();
    l.lock();
    try {
      channels.clear();
    } finally {
      l.unlock();
    }
```

This way, no I/O under lock (blocking operations happen outside the critical section), thread-safety (Channels remain in list during disconnect but are marked as unavailable), consistent state (getChannelById() won't return disconnecting channels) and no deadlock risk (lock is only held for quick memory operations)

norrisjeremy:
Hi @DavidTavoularis,

1. I've dropped the redundant `c.session == this` check in `getChannelById()`.
2. I've changed to using `channels.remove(c)` instead of an iterator in `delChannel()`.
3. I elected to use an iterator so that each channel gets removed from the list even if the `disconnect()` method happened to throw an Exception. I.e., if the list contains three channels and we successfully disconnect the first two, but the last one throws an Exception, all three would remain in the list instead of just the last one.
4. I added an underscore to the internal `_disconnect()` method as it is a common strategy we already use to denote internal use methods that aren't meant for public consumption (which are typically wrapped by a non underscore leading method name that is part of the public API). For example, see the `write() / _write()` and `setPortForwardingR() / _setPortForwardingR()` methods in the Session class for other examples.
5. I'm also not sure that I'm fully comfortable with the idea of copying the list, dropping the lock & then reacquiring the lock and clearing the channels list as you suggest. That seems to me that it could cause rare race conditions, i.e. if a channel was added after the first lock acquisition but before the second lock acquisition, it wouldn't have it's disconnect() executed.

Thanks,
Jeremy

src/main/java/com/jcraft/jsch/Session.java
norrisjeremy: I've refined this to no longer call the disconnect with the lock acquired.

norrisjeremy:
Hi @DavidTavoularis,

I've further refined this based upon some of your suggestions.
I was able to eliminate the internal-use `_disconnect()` method, and I also refined it to drop the lock before calling `disconnect()` for each channel.
I also changed the `addChannel()` method to return a boolean value to indicate whether the channel was added or not to the `channels` list (based upon whether the session had started it's `disconnect()` sequence).
Let me know what you think.

Thanks,
Jeremy

src/main/java/com/jcraft/jsch/PortWatcher.java
```diff
@@ -205,28 +205,28 @@ void run() {
...
-          channel.setSession(session);
-          channel.init();
-          channel.setInputStream(in);
-          channel.setOutputStream(out);
-          session.addChannel(channel);
-          channel.setSocketPath(socketPath);
-          channel.setOrgIPAddress(socket.getInetAddress().getHostAddress());
-          channel.setOrgPort(socket.getPort());
-          channel.connect(connectTimeout);
+          if (session.addChannel(channel)) {
```
norrisjeremy: The changes here do slightly change the order of operations: the channel is now linked to the session (via the `addChannel()` call) prior to it's `init()` method being executed.
I believe this should be ok based upon a cursory review.

src/main/java/com/jcraft/jsch/Session.java
```diff
@@ -2005,23 +2027,32 @@ void run() {
...
+                }
+                tmp.start();
+              } else {
+                packet.reset();
+                buf.putByte((byte) SSH_MSG_CHANNEL_OPEN_FAILURE);
+                buf.putInt(buf.getInt());
+                buf.putInt(Channel.SSH_OPEN_RESOURCE_SHORTAGE);
+                buf.putString(Util.empty);
+                buf.putString(Util.empty);
+                write(packet);
```
norrisjeremy, line 2059: I added this to send a channel open failure back with a resource shortage reason code in case the channel could not be created because the session has started it's `disconnect()` sequence.

src/main/java/com/jcraft/jsch/Session.java
```diff
@@ -2544,11 +2604,14 @@ public int setPortForwardingR(String conf) throws JSchException {
...
-    channel.setHost(host);
-    channel.setPort(port);
-    return channel;
+    if (this.addChannel(channel)) {
+      channel.init();
+      channel.setHost(host);
+      channel.setPort(port);
+      return channel;
+    } else {
+      throw new JSchException("Unable to add channel");
```
norrisjeremy: I added this to throw an exception in case the session's `disconnect()` sequence has started and the channel cannot be added.

src/main/java/com/jcraft/jsch/Session.java
```diff
@@ -2140,6 +2193,13 @@ public void disconnect() {
     }
     io = null;
     socket = null;
+    l.lock();
+    try {
+      disconnectingChannels = false;
+      channels.clear();
+    } finally {
+      l.unlock();
+    }
```
norrisjeremy, line 2221: I added this near the end of `disconnect()` after all the IO operations finished and threads stopped, so that the session returns to a blank slate, as users in the past have expressed a desire to reuse session objects.
Without this, the `SessionReconnectIT` would end up failing, since it tests the reuse operation.

norrisjeremy:
Hi @DavidTavoularis,

I applied your suggestions with a few tweaks.
Let me know what you think.

Thanks,
Jeremy

src/main/java/com/jcraft/jsch/Session.java
```diff
@@ -1345,9 +1376,8 @@ Buffer read(Buffer buf) throws Exception {
         buf.rewind();
         buf.getInt();
         buf.getShort();
-        Channel c = Channel.getChannel(buf.getInt(), this);
-        if (c == null) {
-        } else {
+        Channel c = getChannelById(buf.getInt());
+        if (c != null) {
```
norrisjeremy, line 1380: I tweaked this check, as it seemed bizarre to me.

DavidTavoularis:
@norrisjeremy One minor change probably needed:
```
$ git diff
diff --git a/src/main/java/com/jcraft/jsch/Session.java b/src/main/java/com/jcraft/jsch/Session.java
index 6cc1f58..6fdc2e1 100644
--- a/src/main/java/com/jcraft/jsch/Session.java
+++ b/src/main/java/com/jcraft/jsch/Session.java
@@ -2799,11 +2799,11 @@ public class Session {
   }

   boolean addChannel(Channel channel) {
-    channel.setSession(this);
     Lock l = channelsLock.writeLock();
     l.lock();
     try {
       if (!disconnectingChannels) {
+        channel.setSession(this); // Only set session if we can successfully add
         channels.add(channel);
         return true;
       } else {
```

There are also SonarCloud new issues to review

src/main/java/com/jcraft/jsch/Session.java
```diff
@@ -2101,7 +2160,19 @@ public void disconnect() {
...
+    try {
+      disconnectingChannels = true;
+      channelsCopy = new ArrayList<>(channels);
+    } finally {
+      l.unlock();
+    }
+
+    for (Channel c : channelsCopy) {
+      c.disconnect();
+    }
```
norrisjeremy, line 2176: I wonder if we ought to put the channel disconnect into a try / catch block to ignore any exceptions?
That way even if the call to disconnect a particular channel failed, we would continue to proceed with disconnecting the other channels and then the session itself.

src/main/java/com/jcraft/jsch/Session.java
```diff
@@ -2101,7 +2160,26 @@ public void disconnect() {
...
+
+    for (Channel c : channelsCopy) {
+      try {
+        c.disconnect();
+      } catch (Exception e) {
+        if (getLogger().isEnabled(Logger.DEBUG)) {
+          getLogger().log(Logger.DEBUG, "Exception disconnecting channel of type "
+              + Util.byte2str(c.type) + ": " + e.getMessage(), e);
+        }
```
norrisjeremy, line 2174: I've tweaked this to catch any exceptions while disconnecting individual channels so that it doesn't prematurely halt disconnection of the session itself.

src/main/java/com/jcraft/jsch/Session.java
```diff
@@ -2087,6 +2134,18 @@ void run() {
     isConnected = false;
   }
 
+  void delChannel(Channel c) {
+    Lock l = channelsLock.writeLock();
+    l.lock();
+    try {
+      if (!disconnectingChannels) {
+        channels.remove(c);
+      }
```
norrisjeremy: @DavidTavoularis I'm thinking that we should revert the part of a73a68e that predicated the removal of the channel on `!disconnectingChannels` here.

This is because we've now introduced the possibility that a user could call `Session.disconnect()`, and then an individual channel could be disconnected, but still remain in the `channels` list.
Concurrently, it's then possible for a message to come in for a particular channel from the remote that we don't handle until after the `Channel.disconnect()` method has completed (since `Session::run` is driven by a different thread), and then it's `Session.getChannelById()` could still return this disconnected Channel object and thus lead to unexpected post-disconnection handling of the received message

Functionally, I don't believe there is any harm from always removing the channel from the channels list, even if the session disconnection routine has started.

DavidTavoularis:
> One concern that I still need to think over with regards to this more radical approach from this PR is the possibility for any new memory leaks. Now that both the Channel and Session objects both reference each other, it might be possible for a user to cause a memory leak if they fail to disconnect either the Channel or Session.

@norrisjeremy  Regarding the memory leak risk scenarios:
  1. User creates session, opens channels, then abandons all references without calling disconnect
    - Session holds strong references to channels in List<Channel> channels
    - Each channel holds strong reference to session via this.session
    - Neither can be garbage collected despite being unreachable
  2. User calls channel.disconnect() but not session.disconnect() -> Channel removes itself via delChannel() and nulls this.session
  3. User calls session.disconnect() but not individual channel disconnects -> Session calls disconnect() on all channels

 The original static pool also had leak potential, but differently:
  - Static Vector<Channel> pool held all channels globally
  - Channels could only be removed by explicit cleanup
  - But at least sessions could be GC'd independently

A possible solution would be the WeakReference approach in Channel:
  private WeakReference<Session> sessionRef;
But I don't like it...

norrisjeremy:
Hi @DavidTavoularis,

I made a small update to `Session.addChannel()` to also check that `Session.isConnected`.
Otherwise, there was possibility that `Session.addChannel()` could succeed while after `Session.disconnect()` flipped the `Session.disconnectingChannels` flag back off.

Thanks,
Jeremy

norrisjeremy:
Hi @DavidTavoularis,

I've performed some more refactoring to consolidate on using `Session.openChannel()` as the single method for all consumers to create a channel.
Let me know what you think.

Thanks,
Jeremy

src/main/java/com/jcraft/jsch/PortWatcher.java
```diff
@@ -204,29 +204,49 @@ void run() {
...
-          channel.init();
-          channel.setInputStream(in);
-          channel.setOutputStream(out);
-          session.addChannel(channel);
-          channel.setHost(host);
-          channel.setPort(rport);
-          channel.setOrgIPAddress(socket.getInetAddress().getHostAddress());
-          channel.setOrgPort(socket.getPort());
-          channel.connect(connectTimeout);
-          if (channel.exitstatus != -1) {
```
norrisjeremy, line 229: I got rid of this empty check for exitstatus since it never did anything.

norrisjeremy:
> > I've performed some more refactoring to consolidate on using `Session.openChannel()` as the single method for all consumers to create a channel. Let me know what you think.
> 
> @norrisjeremy the refactoring is fine A minor comment: in Session.java, should you just change "private Channel **get**Channel(String type)" to "private Channel **create**Channel(String type)" ?

Done.

PR #934, opened 2025-10-27
Fix logging a 'port is null' error when parsing openssh config file

jjlauer:
OpenSSH config parsing is logging an "error" when the Port is not defined in the openssh config file (which is optional). While it doesn't affect functionality, it looks very important and is considered an ERROR in the logs when using this library. This checks for a null first before logging the error if the port really is bad.


jjlauer:
I know its just a logging statement, but man does this throw serious looking errors in many of our workflows.  Would it be possible to push out a new release soon?

mwiede:
@jjlauer I have released 2.27.6 now

jjlauer:
Much appreciated. Thanks!

Issue #963, opened 2025-12-17
Support OpenSSH-Agent on Windows


bananensplit:
Firstly, **huge thanks for maintaining this library!**

## Main question
Currently, the OpenSSH agent on Windows is not supported. My question is whether there is a way to implement support for this in the near future. From my research (see below), this should not be particularly difficult, as the main logic already exists. One would only need to adapt `SSHAgentConnector.java` or implement a new, separate `AgentConnector`.

## Some technical research
Under **Linux/Unix**, the OpenSSH agent exposes its API using a [Unix domain socket](https://de.wikipedia.org/wiki/Unix_Domain_Socket).
The path of the Unix socket is stored in the environment variable `SSH_AUTH_SOCK`.

Under **Windows**, it exposes the API using a [Windows named pipe](https://learn.microsoft.com/en-us/windows/win32/ipc/named-pipes).
The name of the pipe is fixed as `\\.\pipe\openssh-ssh-agent` and can be accessed more or less like a file.

Both implementations use **exactly the same protocol**, which means that we don’t need to change any underlying logic—only how we actually communicate with the SSH agent.

There already exists an `AgentConnector` that implements support for the SSH agent on Unix: `SSHAgentConnector.java`. This class uses `java.nio.channels.SocketChannel` to connect to the exposed Unix socket and can therefore query the SSH agent. However, this approach does not work on Windows.

## Proposed solution

Create a new `AgentConnector` that uses `java.nio.channels.FileChannel` instead of `java.nio.channels.SocketChannel`. `FileChannel`s are compatible with Windows named pipes. One could open the `FileChannel` like this:
```java
new RandomAccessFile(DEFAULT_OPENSSH_AGENT_NAMED_PIPE, "rw").getChannel();
```
and then use it almost exactly the same way as a `SocketChannel`.

I already have a working implementation that performs very well for my use case:  [https://github.com/bananensplit/jsch/commit/f384f409f51e912c7e967327ff1fdeea0dc09a83](https://github.com/bananensplit/jsch/commit/f384f409f51e912c7e967327ff1fdeea0dc09a83)

I am very eager to contribute, but it would be great to hear from someone with domain-specific knowledge and experience and get feedback on whether this is a good approach.

**P.S.** I am new to open source, so sorry if I missed something important 🙂

norrisjeremy:
Hi @bananensplit,

We merged support for this several releases ago, so I believe this issue can be closed?

Thanks,
Jeremy

PR #964, opened 2025-12-17
Add support for OpenSSH-Agent on Windows

bananensplit:
All relevant Details are contained in https://github.com/mwiede/jsch/issues/963.

### DIFF
--- a/src/main/java/com/jcraft/jsch/FileChannelFactory.java
+++ b/src/main/java/com/jcraft/jsch/FileChannelFactory.java
@@ -4,5 +4,5 @@ import java.io.IOException;
 import java.nio.channels.FileChannel;
 
 public interface FileChannelFactory {
-	FileChannel open(String path) throws IOException;
+  FileChannel open(String path) throws IOException;
 }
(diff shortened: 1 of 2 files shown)


src/main/java/com/jcraft/jsch/FileChannelFactory.java
```diff
@@ -0,0 +1,8 @@
+package com.jcraft.jsch;
+
+import java.io.IOException;
+import java.nio.channels.FileChannel;
+
+public interface FileChannelFactory {
```
norrisjeremy, line 6: I'm a bit confused by the introduction of this interface, as I don't see classes that implement it in this PR?
Am I missing something?

norrisjeremy (commented):
Hi @bananensplit,

Apologies for the extended delay in replying to this PR.

Thanks!
Jeremy

src/main/java/com/jcraft/jsch/FileChannelFactory.java
norrisjeremy: Oh, I see now, you implemented it by using lambda expressions inside `WindowsSSHAgentConnector`.

norrisjeremy (approved):
I'm not a Windows user, but this appears to be simple enough.

PR #922, opened 2025-10-09
fix for infinite looping issue

code-docker:
fix for issue where method falls in to infinite loop when server proposal has consecutive commas or response starts with ,

PR #997, opened 2026-02-18
Infinite loop fix on checkServerExtInfo()

kalithas-sendrayan:
The method **checkServerExtInfo()** goes into an infinite loop when the server proposal contains consecutive commas or when the response starts with a comma.

There was previously an issue in checkServerStrictKex() caused by the same reason and payload, and it has already been fixed. However, after that method is called, checkServerExtInfo() is invoked, which now causes the same issue. Please refer to the pull request below for more details.

Pull request link -> https://github.com/mwiede/jsch/pull/922


kalithas-sendrayan:
Hi @norrisjeremy, a few of the workflows are awaiting approval from the maintainer. Could you please approve them so the build can run and the code can be merged?

PR #1001, opened 2026-02-20
Add FFM version of Pageant AgentConnector

norrisjeremy:
This adds support for using Pageant using the [JEP 454](https://openjdk.org/jeps/454) FFM API with Java 23+ as an alternative to using JNA.

Additionally, include a few minor updates & fixes to the existing JNA based `PageantConnector`.

To use the new `PageantFFMConnector`, users will need to pass an argument of `--enable-native-access=ALL-UNNAMED` (if JSch is on the classpath) or `--enable-native-access=com.jcraft.jsch` (if JSch is on the modulepath).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -284,6 +284,7 @@
                         <function>CreateFileMappingA</function>
                         <function>FindWindowA</function>
                         <function>GetCurrentProcessId</function>
+                        <function>GetCurrentThreadId</function>
                         <function>MapViewOfFile</function>
                         <function>SendMessageA</function>
                         <function>UnmapViewOfFile</function>
--- a/pom.xml
+++ b/pom.xml
@@ -281,23 +281,38 @@
                     <addAsSourceRoot>false</addAsSourceRoot>
                     <functions>
                         <function>CloseHandle</function>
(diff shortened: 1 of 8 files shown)


pom.xml
```diff
@@ -987,5 +1063,62 @@
...
+                        <artifactId>windowsapi-maven-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <goals>
+                                    <goal>windows-api</goal>
+                                </goals>
+                            </execution>
+                        </executions>
+                    </plugin>
+                    <!-- see https://github.com/manuelbl/WindowsApiGenerator/issues/7 -->
```
norrisjeremy: This is a workaround for manuelbl/WindowsApiGenerator#7 that I opened.
I also opened manuelbl/WindowsApiGenerator#9 with a possible solution.
Once this issue is resolved upstream, we should be able to remove this `find-and-replace-maven-plugin`.

norrisjeremy: This has now been resolved with the new 0.8.5 release.

norrisjeremy:
@mwiede I think this is ready to merge now.

PR #1014, opened 2026-03-09
Improve error message when receiving non-EOF status code during read

adrianovalente:
This addresses the issue being discussed at #1013.

The stream returned by `get(String)` currently throws a generic `IOException("error")` when the server returns a non-EOF status code during a read, making it hard to debug the actual cause of the failure.

This PR replaces that generic message with a descriptive one that includes the SFTP status code name.

### Before

```
java.io.IOException: error
```

### After

```
java.io.IOException: SFTP status error: SSH_FX_PERMISSION_DENIED
```

### Changes

- Add a `sshExceptionDescription(int)` helper method that maps SFTP status codes (`SSH_FX_OK`, `SSH_FX_NO_SUCH_FILE`, `SSH_FX_PERMISSION_DENIED`, etc.) to their names.
- Replace `throw new IOException("error")` with `throw new IOException("SFTP status error: " + sshExceptionDescription(i))` in the `read()` method of the `InputStream` returned by `get(String)`.

No behavioral changes. The function still throws a `IOException`, same control flow. Only the exception message is improved.

### DIFF
--- a/src/main/java/com/jcraft/jsch/ChannelSftp.java
+++ b/src/main/java/com/jcraft/jsch/ChannelSftp.java
@@ -2823,16 +2823,26 @@ public class ChannelSftp extends ChannelSession {
 
   private static String sshExceptionDescription(int code) {
     switch (code) {
-      case SSH_FX_OK: return "SSH_FX_OK";
-      case SSH_FX_EOF: return "SSH_FX_EOF";
-      case SSH_FX_NO_SUCH_FILE: return "SSH_FX_NO_SUCH_FILE";
-      case SSH_FX_PERMISSION_DENIED: return "SSH_FX_PERMISSION_DENIED";
-      case SSH_FX_FAILURE: return "SSH_FX_FAILURE";
-      case SSH_FX_BAD_MESSAGE: return "SSH_FX_BAD_MESSAGE";
-      case SSH_FX_NO_CONNECTION: return "SSH_FX_NO_CONNECTION";
-      case SSH_FX_CONNECTION_LOST: return "SSH_FX_CONNECTION_LOST";
-      case SSH_FX_OP_UNSUPPORTED: return "SSH_FX_OP_UNSUPPORTED";
-      default: return "Unknown code: (" + code + ")";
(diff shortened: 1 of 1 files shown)


src/main/java/com/jcraft/jsch/ChannelSftp.java
norrisjeremy: Can we improve this particular exception message as well?

adrianovalente: Sure! I'm on it

src/main/java/com/jcraft/jsch/ChannelSftp.java
```diff
@@ -1423,8 +1423,7 @@ public int read(byte[] d, int s, int len) throws IOException {
               close();
               return -1;
             }
-            // throwStatusError(buf, i);
-            throw new IOException("error");
+            throw new IOException("SFTP status error: " + sshExceptionDescription(i));
```
norrisjeremy: `throwStatusError()` appears to have some logic to try and read a text message included by the server (assuming certain preconditions are met): would it be valuable to try and do something similar here as well?

norrisjeremy: Also, do we think there would be any value in enhancing the `throwStatusError()` method to try and map the SFTP error code number back to it's string form too?
That way there is some amount of symmetry between how this is handled whether we're in a situation in which only an IOException can be thrown and other situations in which an SftpException is thrown?

src/main/java/com/jcraft/jsch/ChannelSftp.java
norrisjeremy: Thank you for submitting a PR as well: it's always helpful when users can contribute improvements.

norrisjeremy (approved):

adrianovalente:
Hi @norrisjeremy, thanks for the review and approval! Quick question - how does the merge/release process usually work on the repo? Also, would it be possible (if it makes sense) to generate a beta or snapshot release with this change so that I can test it in a controlled environment?

PR #1052, opened 2026-05-20
Make Channel and Session implement AutoClosable

chooklii:
I would argue that both Channel and Session should implement AutoCloseable to enable try-with-resources usage.

However, I'm not entirely sure if my suggested implementation is correct—especially for Channel. I can see a scenario where disconnect() should be the function called during the implicit close.

### DIFF
--- a/src/main/java/com/jcraft/jsch/Channel.java
+++ b/src/main/java/com/jcraft/jsch/Channel.java
@@ -33,7 +33,7 @@ import java.io.PipedInputStream;
 import java.io.PipedOutputStream;
 import java.util.concurrent.atomic.AtomicInteger;
 
-public abstract class Channel {
+public abstract class Channel implements AutoCloseable {
 
   static final int SSH_MSG_CHANNEL_OPEN_CONFIRMATION = 91;
   static final int SSH_MSG_CHANNEL_OPEN_FAILURE = 92;
@@ -515,7 +515,7 @@ public abstract class Channel {
    * destination, if possible.
    */
 
-  void close() {
+  public void close() {
     if (close)
       return;
     close = true;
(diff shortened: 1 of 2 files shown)


src/main/java/com/jcraft/jsch/Channel.java
```diff
@@ -515,7 +515,7 @@ void eof() {
    * destination, if possible.
    */
 
-  void close() {
+  public void close() {
```
norrisjeremy, line 518: This isn't correct: the `close()` method here isn't meant for public consumption, as it does not fully perform all necessary cleanup: it is meant as an internal-use only method.
The correct way to dispose of a Channel object is to call `disconnect()`.

src/main/java/com/jcraft/jsch/Session.java
```diff
@@ -204,6 +204,9 @@ public class Session {
     }
   }
 
+  public void close() {
+    this.disconnect();
```
norrisjeremy, line 208: I don't think introducing AutoCloseable semantics to a Session object is correct: there is an expectation that Session objects are re-usable after users call the `disconnect()` method, which doesn't agree with the semantics of `AutoCloseable`, which implys that after an object is fully disposed after calling the `close()` method.

norrisjeremy (changes requested):
As outlined in my comments, changing the `Session` & `Channel` classes to implement `AutoCloseable` isn't correct.

src/main/java/com/jcraft/jsch/Channel.java
norrisjeremy: Also, I believe Channel may have similar semantics as Session, in that it may be re-usable after calling `disconnect()`, which disagrees with the semantics of AutoCloseable.

chooklii:
I closed this PR. As I found this repo only through an PR of an co-worker. My PR was based on the assumption that a channel should be "closed" when disconnecting. 
As this is not the case I can only agree with @norrisjeremy that if disconnecting is not equal to what a try-with would do when closing this should not be implemented this way.

Issue #1045, opened 2026-04-30
OpenSshCertificateAwareIdentityFile.getSignature() ignores negotiated algorithm, always signs with ssh-rsa (SHA-1)


luigidemasi:
## Description

RSA certificate authentication is broken in FIPS-compliant environments due to two related bugs:

1. **Identity matching**: `UserAuthPublicKey._start()` does not map RSA certificate identities to SHA-2 certificate algorithm variants, so the identity is silently skipped during authentication.
2. **Signature algorithm**: `OpenSshCertificateAwareIdentityFile.getSignature(byte[] data, String alg)` ignores the `alg` parameter and always signs with `ssh-rsa` (SHA-1).

## Bug 1: Identity Matching in `UserAuthPublicKey._start()`

Per [draft-miller-ssh-cert-06](https://www.ietf.org/archive/id/draft-miller-ssh-cert-06.txt), `rsa-sha2-256-cert-v01@openssh.com` and `rsa-sha2-512-cert-v01@openssh.com` are protocol negotiation identifiers only — they never appear in the serialized certificate binary. The certificate key type is always `ssh-rsa-cert-v01@openssh.com` for RSA keys.

This means `identity.getAlgName()` always returns `ssh-rsa-cert-v01@openssh.com` for RSA certificate identities. However, the default `PubkeyAcceptedAlgorithms` intentionally excludes `ssh-rsa-cert-v01@openssh.com` (following OpenSSH defaults) and only includes the SHA-2 variants (`rsa-sha2-512-cert-v01@openssh.com`, `rsa-sha2-256-cert-v01@openssh.com`).

In `UserAuthPublicKey._start()`, plain RSA identities (`ssh-rsa`) are correctly mapped to all RSA algorithm variants:

```java
if (_ipkmethod.equals("ssh-rsa")) {
    ipkmethods = new ArrayList<>(rsamethods); // fans out to rsa-sha2-256, rsa-sha2-512, etc.
}
```

But RSA certificate identities have no equivalent mapping. `ssh-rsa-cert-v01@openssh.com` falls through to the `nonrsamethods.contains()` check, which fails because `ssh-rsa-cert-v01@openssh.com` is not in the default `PubkeyAcceptedAlgorithms`. The identity is silently skipped and `getSignature()` is never reached.

## Bug 2: Signature Algorithm in `OpenSshCertificateAwareIdentityFile`

```java
@Override
public byte[] getSignature(byte[] data, String alg) {
[...]

PR #1087, opened 2026-07-15
Accept RSA-CA-signed OpenSSH host certificates (#1085)

### LINKED ISSUE #1085: OpenSSH host certificates signed by an RSA CA are always rejected: CA key algorithm compared to signature algorithm for equality
### Summary

`OpenSshCertificateHostKeyVerifier` rejects **every** OpenSSH host certificate signed by an **RSA CA**, because it compares the CA key's *public key algorithm* against the certificate's *signature algorithm* for string equality. Those are different things for RSA: a `ssh-rsa` CA key signs with `rsa-sha2-512` (or `rsa-sha2-256`) on any modern OpenSSH.

ed25519 and ECDSA CAs are unaffected — for those, the key algorithm and the signature algorithm names happen to be identical, so the equality check accidentally holds.

### The check

`OpenSshCertificateHostKeyVerifier.getSignatureWrapper` (2.28.3):

```java
String signatureAlgorithm = Util.byte2str(signatureBuffer.getString());

if (!caPublicKeyAlgorithm.equals(signatureAlgorithm)) {
  throw new JSchInvalidHostCertificateException(
      "rejected HostKey: signature verification failed, " + "signature algorithm: '"
          + signatureAlgorithm + "' - CA public Key algorithm: '" + caPublicKeyAlgorithm + "'");
}

// Validate that the CA signature algorithm is allowed and available at runtime
session.checkCASignatureAlgorithm(signatureAlgorithm);
```

`caPublicKeyAlgorithm` comes from the certificate's `signature key` field — for an RSA CA that blob's type string is `ssh-rsa`. `signatureAlgorithm` comes from the `signature` field — `rsa-sha2-512`. So the comparison is `"ssh-rsa".equals("rsa-sha2-512")` → `false`, and the certificate is rejected before the signature is ever checked.

### Why it's wrong

Per [RFC 8332 §3](https://datatracker.ietf.org/doc/html/rfc8332#section-3), `rsa-sha2-256` and `rsa-sha2-512` are *signature algorithms* that operate on an unchanged **`ssh-rsa` public key format** — the key type string stays `ssh-rsa`; only the signature algorithm name differs. OpenSSH's `PROTOCOL.certkeys` likewise defines the certificate's `signature key` as the CA's public key and `signature` as a signature made by it, without requiring the two name strings to match.

OpenSSH has signed certificates with `rsa-sha2-512` by default for an RSA CA since 8.2, and accepts exactly this pairing when verifying.

Note this is the same *key-algorithm vs signature-algorithm* conflation as #1045 (`OpenSshCertificateAwareIdentityFile.getSignature()` always signing `ssh-rsa`/SHA-1), which was fixed on the client-authentication side. This is its counterpart on the host-certificate verification side.

### Reproduction

Stock OpenSSH, default flags:

```console
$ ssh-keygen -q -t rsa -b 2048 -f ca -N '' -C 'rsa-ca'
$ ssh-keygen -q -t ed25519 -f hostkey -N '' -C 'host'
$ ssh-keygen -q -s ca -I test-host -h -n host.example.com -V +52w hostkey.pub

$ cut -d' ' -f1 ca.pub
ssh-rsa                                     # <- caPublicKeyAlgorithm

$ ssh-keygen -L -f hostkey-cert.pub | grep 'Signing CA'
        Signing CA: RSA SHA256:6qBbA3B… (using rsa-sha2-512)      # <- signatureAlgorithm
```

Point a `HostKeyRepository` at `@cert-authority * <ca.pub>` and connect to a server presenting `hostkey-cert.pub`:

```
JSchInvalidHostCertificateException: rejected HostKey: signature verification failed,
signature algorithm: 'rsa-sha2-512' - CA public Key algorithm: 'ssh-rsa'
```

The signature is perfectly valid; it is never verified.

(Tested against jsch 2.28.3 with a real Apache MINA sshd presenting an OpenSSH host certificate. Swapping only the CA to ed25519 or ECDSA makes the same setup validate, which isolates the cause to the name comparison rather than to the parsing or the crypto.)

### Suggested fix

The subsequent `session.checkCASignatureAlgorithm(signatureAlgorithm)` already constrains the signature algorithm to the configured `ca_signature_algorithms` allow-list, so the equality check is not what provides the security property — it only needs to establish that the signature algorithm is *usable with the CA key type*:

```java
if (!isSignatureAlgorithmCompatible(caPublicKeyAlgorithm, signatureAlgorithm)) { … }

// ssh-rsa keys may sign with ssh-rsa | rsa-sha2-256 | rsa-sha2-512 (RFC 8332);
// every other key type uses its own name.
static boolean isSignatureAlgorithmCompatible(String caKeyAlgorithm, String signatureAlgorithm) {
  if ("ssh-rsa".equals(caKeyAlgorithm)) {
    return "ssh-rsa".equals(signatureAlgorithm)
        || "rsa-sha2-256".equals(signatureAlgorithm)
        || "rsa-sha2-512".equals(signatureAlgorithm);
  }
  return caKeyAlgorithm.equals(signatureAlgorithm);
}
```

Happy to send a PR with a test if you'd like it in that shape.

### Environment

- jsch 2.28.3
- Found while wiring `@cert-authority` host-CA trust into [Haven](https://github.com/GlassHaven/Haven) (an Android SSH client), where it means users with an RSA SSH host CA — including the default for some `step-ca` deployments — cannot use host-certificate verification at all and silently fall back to TOFU.

GlassOnTin:
Fixes #1085.

## Problem

`OpenSshCertificateHostKeyVerifier.getSignatureWrapper` compared the CA key's *public-key* algorithm against the certificate's *signature* algorithm for string equality:

```java
if (!caPublicKeyAlgorithm.equals(signatureAlgorithm)) { throw ... }
```

For an RSA CA those legitimately differ. Per [RFC 8332 §3](https://datatracker.ietf.org/doc/html/rfc8332#section-3), an `ssh-rsa` key signs with `rsa-sha2-256`/`rsa-sha2-512` while the key type string stays `ssh-rsa`, and OpenSSH signs host certificates with `rsa-sha2-512` by default since 8.2. So `"ssh-rsa".equals("rsa-sha2-512")` is `false` and **every host certificate signed by an RSA CA was rejected before its (valid) signature was ever verified**. ed25519/ECDSA CAs were unaffected only because their key and signature algorithm names coincide.

This is the host-verification counterpart of the same key-alg-vs-sig-alg conflation as #1045 (client side, already fixed).

## Fix

Replace the equality check with `isSignatureAlgorithmCompatible`, which accepts `ssh-rsa | rsa-sha2-256 | rsa-sha2-512` for an `ssh-rsa` CA key and requires an exact match for every other key type. The subsequent `session.checkCASignatureAlgorithm(...)` still enforces the `ca_signature_algorithms` allow-list, so the equality check was not what provided the security property.

## Test

`OpenSshCertificateHostKeyVerifierTest.testCheckSignature_rsaCaSigningWithRsaSha2_isAccepted` parses a **real** OpenSSH host certificate signed by an RSA CA with `rsa-sha2-512` (new fixtures under `src/test/resources/certificates/rsa_host_ca/`, generated with stock `ssh-keygen`) and asserts `checkSignature` verifies it.

Verified both directions on JDK 25:

- **Before the fix** (equality check restored), the test fails with exactly the reported error:
  ```
  JSchInvalidHostCertificateException: rejected HostKey: signature verification failed,
  signature algorithm: 'rsa-sha2-512' - CA public Key algorithm: 'ssh-rsa'
  ```
- **After the fix**, the real signature verifies and the test passes.

A small companion test, `testIsSignatureAlgorithmCompatible`, documents the accepted pairings and confirms non-RSA CA keys still require an exact name match.

Full `OpenSshCertificate*Test` + `KnownHostsTest` unit set: 221 pass, 0 failures.

The fixture directory contains only the host `*-cert.pub` and the CA public key `ca.pub` (so the RSA CA type can be re-verified with `ssh-keygen -L`); no private keys are committed.


norrisjeremy (approved):

GlassOnTin:
CI is green — all six `test` matrix jobs (JDK 8, 11, 17, 21, 25, 26) pass, along with `build` and SonarCloud. The JDK runs include the Docker `HostCertificateIT` suite, so the existing ed25519/ECDSA CA paths are unaffected by the change.

PR #1098, opened 2026-07-28
Fix #1091: Parse certificate-form revoked host keys

### LINKED ISSUE #1091: Cert Revocation Type String Bypass
### 1. `@revoked` known_hosts Entries Written Using a Certificate Type String Are Silently Dropped

**Severity:** Medium
**CVSS Score:** 5.9 — CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:H/A:N
**Location:** Parser gap: `src/main/java/com/jcraft/jsch/KnownHosts.java:181-187` (`name2type`/`addInvalidLine`), `:270-273` (`addInvalidLine`), `:397` (`getHostKey` UNKNOWN-exclusion); recognized-type table: `src/main/java/com/jcraft/jsch/HostKey.java:33-36`; default negotiable algorithm list: `src/main/java/com/jcraft/jsch/JSch.java:50-51`
**Component/Version:** `com.github.mwiede:jsch` 2.28.4
**CWE:** CWE-299: Improper Check for Certificate Revocation
**Affected Component:** `known_hosts` parsing (`KnownHosts`) as consumed by OpenSSH host-certificate revocation checking (`OpenSshCertificateUtil.getRevokedKeys()`)

---

#### Description

JSch's `known_hosts` parser recognizes host-key type tokens via `HostKey.name2type()`, whose table is a small, hard-coded list:

```java
// HostKey.java:33-36
private static final byte[][] names =
    {Util.str2byte("ssh-dss"), Util.str2byte("ssh-rsa"), Util.str2byte("ecdsa-sha2-nistp256"),
        Util.str2byte("ecdsa-sha2-nistp384"), Util.str2byte("ecdsa-sha2-nistp521"),
        Util.str2byte("ssh-ed25519"), Util.str2byte("ssh-ed448")};
```

This table contains **no OpenSSH certificate type strings** (`*-cert-v01@openssh.com`). Yet JSch's own **default** `server_host_key` negotiation list treats these certificate types as first-class, negotiable host-key algorithms:

```java
// JSch.java:50-51
config.put("server_host_key", Util.getSystemProperty("jsch.server_host_key",
    "ssh-ed25519-cert-v01@openssh.com,ecdsa-sha2-nistp256-cert-v01@openssh.com,...,"
    + "ssh-ed25519,ecdsa-sha2-nistp256,...));
```

When `KnownHosts` parses a line whose key-type token is **not** in `HostKey.name2type()`'s table — which includes every certificate type string — it discards the type-binding entirely and stores the *whole line* as a marker-less "invalid" entry:

```java
// KnownHosts.java:180-187
if (HostKey.name2type(tmp) != HostKey.UNKNOWN) {
  type = HostKey.name2type(tmp);
} else {
  j = bufl;                                   // forces fall-through below
}
if (j >= bufl) {
  addInvalidLine(Util.byte2str(buf, 0, bufl));  // whole line, no marker preserved
  continue loop;
}
```
```java
// KnownHosts.java:270-273
private void addInvalidLine(String line) throws JSchException {
  HostKey hk = new HostKey(line, HostKey.UNKNOWN, null);  // no marker set
  pool.addElement(hk);
}
```

Both the lost `@revoked` marker and the `type == HostKey.UNKNOWN` value cause this entry to be excluded from `KnownHosts.getHostKey()` (`KnownHosts.java:397: if (hk.type == HostKey.UNKNOWN) continue;`), which is the exact method `OpenSshCertificateUtil.getRevokedKeys()` uses to build the revoked-key set consulted during certificate verification. **The end result: an administrator who writes `@revoked <host> ssh-ed25519-cert-v01@openssh.com <cert-blob>` into `known_hosts` — a natural choice, since JSch itself treats that string as a normal algorithm name everywhere else — has that revocation entry silently discarded at parse time, and a host presenting that exact (revoked) certificate is never checked against it.**

This is specifically a parser/type-table gap, not a flaw in the revocation *comparison* logic itself: `OpenSshCertificateHostKeyVerifier`'s actual revocation check compares the certificate's embedded base public key against revoked entries authored using the **base** key type (e.g. `ssh-ed25519`) — those work correctly. The vulnerability is that JSch provides no normalization of a cert-type token to its base type before the `name2type()` lookup, and does not document that revoked-certificate entries must be authored with the base type rather than the (equally valid-looking, JSch-recognized-elsewhere) certificate type string.

---

#### Steps to Reproduce (Dynamic, Real-Environment PoC)

Reproduced with a **genuine, validly-signed OpenSSH host certificate** (real `ssh-keygen -s`), presented over a real SSH2 handshake, with two `known_hosts` variants differing only in how the `@revoked` entry's type field is written — providing a clean positive/negative comparison.

**Environment:** macOS, OpenJDK 25, JSch 2.28.4, OpenSSH `ssh-keygen`, Python 3.14 + `paramiko==5.0.0`.

**1. Generate CA, host key, and a real certificate** (same infrastructure as the companion cert-fallback finding):
```bash
ssh-keygen -t ed25519 -f ca_key -N ""
ssh-keygen -t ed25519 -f host_key -N ""
ssh-keygen -s ca_key -h -I host01 -n 127.0.0.1 -V always:forever host_key.pub
```

**2. Build two `known_hosts` files**, both trusting the CA, differing only in the `@revoked` line's type token:

```bash
CA_LINE=$(cat ca_key.pub | awk '{print "* "$1" "$2}')
CERT_TYPE_LINE=$(cat host_key-cert.pub | awk '{print $1" "$2}')   # ssh-ed25519-cert-v01@openssh.com <cert-blob>
BASE_LINE=$(cat host_key.pub | awk '{print $1" "$2}')             # ssh-ed25519 <plain-pubkey>

# vulnerable: admin revokes using the CERTIFICATE's own type string
echo "@cert-authority $CA_LINE" > known_hosts_revoked_certtype
echo "@revoked 127.0.0.1 $CERT_TYPE_LINE" >> known_hosts_revoked_certtype

# control: admin revokes using the BASE key type (extracted from the cert)
echo "@cert-authority $CA_LINE" > known_hosts_revoked_basetype
echo "@revoked 127.0.0.1 $BASE_LINE" >> known_hosts_revoked_basetype
```

**3. JSch client (`CertRevocationBypassPoc.java`), `StrictHostKeyChecking=yes`, `host_certificate_to_key_fallback=no`** (isolating this purely to the revocation-check path, independent of the companion fallback finding):

```java
jsch.setKnownHosts(knownHostsPath);
Session session = jsch.getSession("victim", "127.0.0.1", port);
session.setConfig("StrictHostKeyChecking", "yes");
session.setConfig("server_host_key", "ssh-ed25519-cert-v01@openssh.com,ssh-ed25519");
session.setConfig("host_certificate_to_key_fallback", "no");
session.connect(10000);
```

**4. Run against the SAME real certificate-presenting server, swapping only the `known_hosts` file:**

```
=== TEST 1: revoked using CERT-TYPE STRING ===
[client:CERT-TYPE-REVOKED] connecting -- known_hosts marks this host's certificate as revoked...
[client:CERT-TYPE-REVOKED] session.connect() SUCCEEDED
[client:CERT-TYPE-REVOKED] RESULT: VULNERABLE -- a revoked certificate was silently accepted.

=== TEST 2 (control): revoked using BASE-TYPE STRING ===
[client:BASE-TYPE-REVOKED-control] connecting -- known_hosts marks this host's certificate as revoked...
[client:BASE-TYPE-REVOKED-control] session.connect() FAILED (correctly rejected):
com.jcraft.jsch.JSchRevokedHostKeyException: Rejected certificate 'host01': host certificate
public key is marked as revoked for [127.0.0.1]:8042
[client:BASE-TYPE-REVOKED-control] RESULT: revocation enforced correctly.
```

The identical certificate, the identical server, the identical `@cert-authority` trust configuration — the **only** difference is whether the admin wrote the `@revoked` line's type field as the certificate type or the base type. One is silently ignored; the other correctly and hard-fails the connection with `JSchRevokedHostKeyException`.

---

#### Impact

- **Confidentiality:** None directly.
- **Integrity:** High — an intended, explicitly-authored security control (host key/certificate revocation) silently does not take effect, with **no error, warning, or log message indicating the entry was rejected as unparseable**. An administrator who revokes a compromised host's certificate in the natural way (using the certificate's own type string, which JSch treats as a normal algorithm name everywhere else in its configuration) will believe the host is blocked when it is not.
- **Availability:** None.
- **Preconditions:** the deployment must use OpenSSH certificate-based host authentication (a non-default but supported and documented JSch feature) and the revoking administrator must author the `@revoked` entry using the certificate-type token rather than the base key type — a very plausible, arguably more intuitive authoring choice that JSch's own configuration surface (treating cert types as ordinary algorithm names) actively encourages, and which is not warned against anywhere in JSch's documentation or code.

---

#### Proof of Concept — Full Script

`CertRevocationBypassPoc.java`:

```java
import com.jcraft.jsch.*;

public class CertRevocationBypassPoc {
  public static void main(String[] args) throws Exception {
    int port = Integer.parseInt(args[0]);
    String knownHostsPath = args[1];
    String label = args[2];

    JSch jsch = new JSch();
    jsch.setKnownHosts(knownHostsPath);

    Session session = jsch.getSession("victim", "127.0.0.1", port);
    session.setConfig("StrictHostKeyChecking", "yes");
    session.setConfig("server_host_key", "ssh-ed25519-cert-v01@openssh.com,ssh-ed25519");
    session.setConfig("host_certificate_to_key_fallback", "no");
    session.setPassword("anypassword");

    try {
      session.connect(10000);
      System.out.println("[client:" + label + "] session.connect() SUCCEEDED -- VULNERABLE if cert-type entry");
    } catch (Throwable t) {
      System.out.println("[client:" + label + "] session.connect() FAILED (correctly rejected): " + t);
    } finally {
      session.disconnect();
    }
  }
}
```

`malicious_server.py` reuses the `cert-fallback` mode's `load_cert_host_key()` helper (see the companion certificate-fallback report) to present the genuine certificate over the transport; no additional server-side logic is needed since the vulnerability is entirely in client-side `known_hosts` parsing.

**Run:**
```bash
ssh-keygen -t ed25519 -f ca_key -N ""
ssh-keygen -t ed25519 -f host_key -N ""
ssh-keygen -s ca_key -h -I host01 -n 127.0.0.1 -V always:forever host_key.pub

CA_LINE=$(cat ca_key.pub | awk '{print "* "$1" "$2}')
echo "@cert-authority $CA_LINE" > known_hosts_revoked_certtype
echo "@revoked 127.0.0.1 $(cat host_key-cert.pub | awk '{print $1" "$2}')" >> known_hosts_revoked_certtype
echo "@cert-authority $CA_LINE" > known_hosts_revoked_basetype
echo "@revoked 127.0.0.1 $(cat host_key.pub | awk '{print $1" "$2}')" >> known_hosts_revoked_basetype

python3 malicious_server.py cert-fallback 8041 &
javac -cp jsch-2.28.4.jar CertRevocationBypassPoc.java
java  -cp jsch-2.28.4.jar:bcprov-jdk18on-1.84.jar:. CertRevocationBypassPoc 8041 known_hosts_revoked_certtype VULN

python3 malicious_server.py cert-fallback 8042 &
java  -cp jsch-2.28.4.jar:bcprov-jdk18on-1.84.jar:. CertRevocationBypassPoc 8042 known_hosts_revoked_basetype CONTROL
```

---

#### Remediation

1. Normalize a certificate-type token to its base algorithm name (`OpenSshCertificateKeyTypes.getBaseKeyType()` already exists in the codebase for this exact purpose but is never invoked during `known_hosts` parsing) before calling `HostKey.name2type()`, so `@revoked`/`@cert-authority` lines authored with either the cert type or the base type parse identically.
2. Preserve the marker and raw key bytes for any line whose type is unrecognized, rather than collapsing it to a marker-less `UNKNOWN` entry — at minimum, log a loud warning when a `@revoked`-marked line is dropped due to an unrecognized type, so the administrator's mistake is visible rather than silent.
3. Build `OpenSshCertificateUtil.getRevokedKeys()`'s comparison set by scanning the raw `known_hosts` pool for `@revoked` markers directly (matching on raw key bytes) rather than exclusively through `getHostKey()`'s type-filtered lookup.

---

#### References

- CWE-299: Improper Check for Certificate Revocation
- OpenSSH `PROTOCOL.certkeys` / `sshd(8)` `@revoked`, `@cert-authority` `known_hosts` marker semantics
- Source: `src/main/java/com/jcraft/jsch/KnownHosts.java`, `HostKey.java`, `JSch.java`, `OpenSshCertificateUtil.java` (JSch 2.28.4, `com.github.mwiede:jsch`)

luigidemasi:
## Summary

- parse supported certificate-form `@revoked` entries into their embedded base public keys
- validate the declared and embedded key families, including RSA SHA-2 aliases
- preserve malformed or mismatched certificate entries without interrupting later parsing
- add regression coverage for Ed25519, ECDSA, RSA SHA-2, and invalid certificate records

Fixes #1091

## Testing

- `JAVA_HOME=/usr/lib/jvm/java-25-openjdk ./mvnw -B clean install`
- 478 tests passed

### DIFF
--- a/src/main/java/com/jcraft/jsch/KnownHosts.java
+++ b/src/main/java/com/jcraft/jsch/KnownHosts.java
@@ -252,24 +252,13 @@ class KnownHosts implements HostKeyRepository {
         // System.err.println(host);
         // System.err.println("|"+key+"|");
 
-        byte[] keyData;
         if (revokedCertificate) {
-          try {
-            OpenSshCertificate certificate = OpenSshCertificateParser.parse(jsch.instLogger,
-                Util.fromBase64(Util.str2byte(key), 0, key.length()));
-            keyData = certificate.getCertificatePublicKey();
-            type = HostKey.name2type(Util.byte2str(new Buffer(keyData).getString()));
-            if (type == HostKey.UNKNOWN || type != getCertificateBaseType(keyType)) {
-              throw new JSchException("Certificate key type does not match known_hosts key type");
-            }
-          } catch (JSchException | RuntimeException e) {
-            addInvalidLine(Util.byte2str(buf, 0, bufl));
-            continue loop;
-          }
-        } else {
-          keyData = Util.fromBase64(Util.str2byte(key), 0, key.length());
+          addRevokedCertificateHostKey(Util.byte2str(buf, 0, bufl), marker, host, keyType, key,
+              comment);
+          continue loop;
         }
 
+        byte[] keyData = Util.fromBase64(Util.str2byte(key), 0, key.length());
         pool.addElement(new HashedHostKey(marker, host, type, keyData, comment));
       }
       if (error) {
@@ -285,6 +274,28 @@ class KnownHosts implements HostKeyRepository {
   private void addInvalidLine(String line) throws JSchException {
     HostKey hk = new HostKey(line, HostKey.UNKNOWN, null);
     pool.addElement(hk);
+  }
+
+  // Unnamed catch variables require Java 22, but this source set targets Java 8.
+  @SuppressWarnings("java:S7467")
+  private void addRevokedCertificateHostKey(String line, String marker, String host, String keyType,
+      String key, String comment) throws JSchException {
+    byte[] keyData;
+    int type;
+    try {
+      OpenSshCertificate certificate = OpenSshCertificateParser.parse(jsch.instLogger,
+          Util.fromBase64(Util.str2byte(key), 0, key.length()));
+      keyData = certificate.getCertificatePublicKey();
+      type = HostKey.name2type(Util.byte2str(new Buffer(keyData).getString()));
+      if (type == HostKey.UNKNOWN || type != getCertificateBaseType(keyType)) {
+        throw new JSchException("Certificate key type does not match known_hosts key type");
+      }
+    } catch (JSchException | RuntimeException ignored) {
+      addInvalidLine(line);
+      return;
+    }
+
+    pool.addElement(new HashedHostKey(marker, host, type, keyData, comment));
   }
 
   private static int getCertificateBaseType(String keyType) {
--- a/src/main/java/com/jcraft/jsch/KnownHosts.java
+++ b/src/main/java/com/jcraft/jsch/KnownHosts.java
@@ -169,7 +169,6 @@ class KnownHosts implements HostKeyRepository {
         }
 
         sb.setLength(0);
-        type = -1;
         while (j < bufl) {
           i = buf[j++];
           if (i == 0x20 || i == '\t') {
@@ -177,10 +176,11 @@ class KnownHosts implements HostKeyRepository {
           }
           sb.append((char) i);
         }
-        String tmp = sb.toString();
-        if (HostKey.name2type(tmp) != HostKey.UNKNOWN) {
-          type = HostKey.name2type(tmp);
-        } else {
+        String keyType = sb.toString();
+        boolean revokedCertificate =
+            "@revoked".equals(marker) && OpenSshCertificateKeyTypes.isCertificateKeyType(keyType);
+        type = HostKey.name2type(keyType);
+        if (type == HostKey.UNKNOWN && !revokedCertificate) {
           j = bufl;
         }
         if (j >= bufl) {
@@ -252,10 +252,25 @@ class KnownHosts implements HostKeyRepository {
         // System.err.println(host);
         // System.err.println("|"+key+"|");
 
-        HostKey hk = null;
-        hk = new HashedHostKey(marker, host, type,
-            Util.fromBase64(Util.str2byte(key), 0, key.length()), comment);
-        pool.addElement(hk);
+        byte[] keyData;
+        if (revokedCertificate) {
+          try {
+            OpenSshCertificate certificate = OpenSshCertificateParser.parse(jsch.instLogger,
+                Util.fromBase64(Util.str2byte(key), 0, key.length()));
+            keyData = certificate.getCertificatePublicKey();
+            type = HostKey.name2type(Util.byte2str(new Buffer(keyData).getString()));
+            if (type == HostKey.UNKNOWN || type != getCertificateBaseType(keyType)) {
+              throw new JSchException("Certificate key type does not match known_hosts key type");
+            }
+          } catch (JSchException | RuntimeException e) {
+            addInvalidLine(Util.byte2str(buf, 0, bufl));
+            continue loop;
+          }
+        } else {
+          keyData = Util.fromBase64(Util.str2byte(key), 0, key.length());
+        }
+
+        pool.addElement(new HashedHostKey(marker, host, type, keyData, comment));
       }
       if (error) {
         throw new JSchException("KnownHosts: invalid format");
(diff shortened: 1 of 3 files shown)


norrisjeremy (approved):
Hi @luigidemasi,

Thank you for investigating #1091 and submitting a PR with a resolution!
I've confirmed that the example PoC provided in #1091 correctly rejects connections to a server with a revoked cert-type key line.

Thanks,
Jeremy

PR #1100, opened 2026-07-29
Fix #1099: Respect known_hosts markers

### LINKED ISSUE #1099: @cert-authority entries cause false CHANGED results and can be deleted on replacement
## Description

`KnownHosts.check()` compares matching host/type entries without considering the `HostKey` marker. As a result, a valid `@cert-authority` entry is treated as an ordinary trusted host key.

When a server presents a different plain host key of the same algorithm, `check()` returns `CHANGED` instead of `NOT_INCLUDED`. This sends `Session` through the false “REMOTE HOST IDENTIFICATION HAS CHANGED” path.

If the user accepts replacement with `StrictHostKeyChecking=ask`, `Session` calls `remove(host, type, null)`. That operation is also marker-blind and deletes the matching `@cert-authority` entry before adding the plain host key, removing the configured CA trust anchor.

## Steps to Reproduce

1. Load `src/test/resources/certificates/known_hosts`, which contains:

   ```
   @cert-authority localhost ssh-ed25519 <CA-public-key>
   ```

2. Read the different plain Ed25519 host key from:

   ```
   src/test/resources/certificates/host/ssh_host_ed25519_key.pub
   ```

3. Call:

   ```java
   repository.check("localhost", hostKeyBlob);
   ```

4. Optionally call the same removal used by the accepted changed-key path:

   ```java
   repository.remove("localhost", "ssh-ed25519", null);
   ```

## Expected Behavior

- The CA entry should not participate in ordinary plain host-key matching.
- With no ordinary host-key entry, `check()` should return `NOT_INCLUDED`.
- Replacing an ordinary host key should not remove the `@cert-authority` entry.

## Actual Behavior

- `check()` returns `CHANGED`.
- A connection reports a false host-key-change/MITM warning.
- Accepting replacement removes the `@cert-authority` entry.

## Additional Context

`KnownHosts.check()` compares only the host, key type, and key bytes. `KnownHosts.remove()` similarly removes matching entries without considering their markers.

A fix must preserve the special handling of `@revoked`: a matching revoked entry currently returns `OK`, after which `Session` performs the marker-specific revocation check. Therefore, simply ignoring every marked entry in `check()` would break revocation handling.

Suggested regression coverage belongs in `KnownHostsTest`:

1. A different plain key returns `NOT_INCLUDED` when only an `@cert-authority` entry exists.
2. The CA key bytes are not accepted as an ordinary plain host key.
3. Ordinary host-key replacement preserves the `@cert-authority` entry.

This is independent of #1091 and requires separate handling.

luigidemasi:
## Summary

- exclude certificate-authority and unknown marked entries from ordinary host-key matching
- preserve exact revoked-key matches so Session can enforce revocation without false changed-key results
- retain marked entries when replacing an ordinary host key
- normalize null markers to the canonical unmarked representation
- document removal semantics and add regression coverage for marker handling

Fixes #1099

## Testing

- `JAVA_HOME=/usr/lib/jvm/java-25-openjdk ./mvnw -B -Dtest=HostKeyTest,KnownHostsTest test`
- `JAVA_HOME=/usr/lib/jvm/java-25-openjdk ./mvnw -B clean install`
- 477 tests passed

## Notes

- based directly on `origin/master`
- independent of #1098

Generated by Codex via /oss-address-review

### DIFF
--- a/src/main/java/com/jcraft/jsch/HostKey.java
+++ b/src/main/java/com/jcraft/jsch/HostKey.java
@@ -65,7 +65,7 @@ public class HostKey {
 
   public HostKey(String marker, String host, int type, byte[] key, String comment)
       throws JSchException {
-    this.marker = marker;
+    this.marker = marker == null ? "" : marker;
     this.host = host;
     if (type == GUESS) {
       if (key[8] == 'd') {
--- a/src/main/java/com/jcraft/jsch/HostKeyRepository.java
+++ b/src/main/java/com/jcraft/jsch/HostKeyRepository.java
@@ -51,15 +51,25 @@ public interface HostKeyRepository {
   void add(HostKey hostkey, UserInfo ui);
 
   /**
-   * Removes a host key if there exists mached key with <code>host</code>, <code>type</code>.
+   * Removes host keys matching <code>host</code> and <code>type</code>.
+   *
+   * <p>
+   * When both arguments are non-null, entries with a non-empty marker are preserved. A null
+   * <code>host</code> clears the repository, while a null <code>type</code> removes all keys
+   * matching <code>host</code>, including marked entries.
    *
    * @see #remove(String host, String type, byte[] key)
    */
   void remove(String host, String type);
 
   /**
-   * Removes a host key if there exists a matched key with <code>host</code>, <code>type</code> and
-   * <code>key</code>.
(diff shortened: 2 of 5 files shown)


norrisjeremy (approved):
Hi @luigidemasi,

Thank you for identifying this issue and submitting a PR to resolve it!

Thanks,
Jeremy

PR #1101, opened 2026-07-29
Add Session accessors for the negotiated algorithms

fornwall:
Adds accessors to `Session` for the algorithms negotiated during key exchange, one per proposal and direction:

```java
public String getKexAlgorithm()
public String getServerHostKeyAlgorithm()
public String getCipherAlgorithmC2S()
public String getCipherAlgorithmS2C()
public String getMacAlgorithmC2S()
public String getMacAlgorithmS2C()
public String getCompressionAlgorithmC2S()
public String getCompressionAlgorithmS2C()
```

Until now the only way to observe the negotiation outcome was to parse the `kex: ...` INFO log lines. Having accessors is useful for displaying or auditing connection details, and for asserting that configured algorithm preferences are honoured.

The negotiation result is published to a new volatile `negotiatedAlgorithms` field when `receive_newkeys()` puts the new keys into use. The accessors therefore return `null` until the first key exchange has completed, keep reporting the algorithms currently in effect while a rekey is in progress, and never report a key exchange that failed. The volatile write safely publishes the array to callers on other threads. The MAC accessors return `null` for a direction whose negotiated cipher is AEAD, since no separate MAC is used then.

### DIFF
--- a/src/main/java/com/jcraft/jsch/Session.java
+++ b/src/main/java/com/jcraft/jsch/Session.java
@@ -3237,6 +3237,24 @@ public class Session {
     return getNegotiatedAlgorithm(KeyExchange.PROPOSAL_COMP_ALGS_STOC);
   }
 
+  /**
+   * Returns whether strict key exchange was negotiated with the server.
+   *
+   * <p>
+   * Strict key exchange is only negotiated during the initial key exchange, so unlike the
+   * negotiated algorithms this value does not change when the session is rekeyed. It returns
+   * {@code false} again once the session is disconnected.
+   *
+   * @return {@code true} if strict key exchange, the countermeasure against
+   *         <a href="https://nvd.nist.gov/vuln/detail/CVE-2023-48795">CVE-2023-48795</a>, was
+   *         negotiated, or {@code false} if the initial key exchange has not started yet, if the
+   *         server did not support it, or if it was turned off with the {@code enable_strict_kex}
+   *         configuration
+   */
+  public boolean isStrictKex() {
+    return doStrictKex;
+  }
+
   public void sendIgnore() throws Exception {
     Buffer buf = new Buffer();
     Packet packet = new Packet(buf);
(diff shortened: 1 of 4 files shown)


norrisjeremy:
Hi @fornwall,

This is an interesting proposal.
Do you think there would be any value in adding a method that could be used to also determine whether `kex-strict` was negotiated as well?
I believe this could be accomplished as a simple wrapper around `doStrictKex`.

Thanks,
Jeremy

src/main/java/com/jcraft/jsch/Session.java
```diff
@@ -103,6 +103,7 @@ public class Session {
   private int seqo = 0;
 
   String[] guess = null;
+  volatile String[] negotiatedAlgorithms = null;
```
norrisjeremy, line 106: Any reason for this to be package private instead of private?

fornwall: @norrisjeremy For the test to set the field.

I personally think that is cleanest, but let me know if you want me to switch to reflection instead.

norrisjeremy: Very good, I overlooked that, thanks!

norrisjeremy: No need to make any changes, it's fine to have it package private to allow it to be manipulated for unit tests.

fornwall:
> Do you think there would be any value in adding a method that could be used to also determine whether kex-strict was negotiated as well?
I believe this could be accomplished as a simple wrapper around doStrictKex.

👍 Pushed https://github.com/mwiede/jsch/pull/1101/changes/e078171d28516d697ea9563725238388ce2b920c with that.

norrisjeremy (approved):
Hi @fornwall,

I am good with adding these new methods.
Thank you for your contribution!

Thanks,
Jeremy

fornwall:
Thanks yourself for the quick revew and feedback!

PR #1117, opened 2026-08-10
Report a refused agent identity removal instead of always succeeding

fornwall:
Fix forwarded-agent removal replies so the remote caller sees the repository's actual result.

### Changes

- Reply with `SSH_AGENT_SUCCESS` or `SSH_AGENT_FAILURE` based on `IdentityRepository.remove(byte[])`.
- Add `IdentityRepository.removeAllIdentities()`, a default method that calls the existing `removeAll()` and reports success. Existing implementations keep their current behavior, while agent-backed repositories can return the agent's real response.
- Return the remove-all response already read by `AgentProxy` and propagate it through `AgentIdentityRepository`.
- Remove identities from `IdentityRepositoryWrapper`'s cache as well as its delegate.

The remove-all result is reported directly rather than inferred from `getIdentities()`: a locked or unreachable agent may also fail identity listing, making a failed removal look like an empty repository.

### Tests

Added wire-level coverage for successful, refused, and unreachable removals, plus compatibility coverage for existing `removeAll()` implementations and overrides. Added wrapper tests for cached and delegated identities.

`./mvnw verify -DskipITs=true`: 500 tests, 0 failures.

Disclaimer: Created with claude code and opus 5. Reviewed and iterated on by me.

### DIFF
--- a/src/main/java/com/jcraft/jsch/AgentIdentityRepository.java
+++ b/src/main/java/com/jcraft/jsch/AgentIdentityRepository.java
@@ -31,6 +31,7 @@ import java.util.Vector;
 public class AgentIdentityRepository implements IdentityRepository {
 
   private AgentProxy agent;
+  private boolean removeAllResult;
 
   public AgentIdentityRepository(AgentConnector connector) {
     this.agent = new AgentProxy(connector);
@@ -52,8 +53,16 @@ public class AgentIdentityRepository implements IdentityRepository {
   }
 
   @Override
-  public void removeAll() {
-    agent.removeAllIdentities();
+  public synchronized void removeAll() {
+    removeAllResult = agent.removeAllIdentities();
+  }
+
+  @Override
+  public synchronized boolean removeAllIdentities() {
+    removeAllResult = true;
+    // Preserve dispatch to removeAll() for subclasses compiled before this method was added.
+    removeAll();
+    return removeAllResult;
   }
 
   @Override
--- a/src/main/java/com/jcraft/jsch/IdentityRepository.java
+++ b/src/main/java/com/jcraft/jsch/IdentityRepository.java
@@ -44,4 +44,21 @@ public interface IdentityRepository {
   public boolean remove(byte[] blob);
 
   public void removeAll();
+
+  /**
+   * Removes all identities and reports whether that succeeded.
+   *
+   * <p>
+   * This is {@link #removeAll()} with an outcome, which a forwarded agent needs in order to answer
+   * an {@code SSH2_AGENTC_REMOVE_ALL_IDENTITIES} request truthfully. Implementations backed by
+   * something that can refuse the removal, such as a locked or unreachable ssh-agent, should
+   * override this and return the real outcome; the default implementation delegates to
+   * {@link #removeAll()} and reports success.
+   *
+   * @return {@code true} if all identities were removed
+   */
+  public default boolean removeAllIdentities() {
+    removeAll();
+    return true;
+  }
 }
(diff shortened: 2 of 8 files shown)


src/main/java/com/jcraft/jsch/AgentIdentityRepository.java
```diff
@@ -52,8 +53,16 @@ public boolean remove(byte[] blob) {
...
+  public synchronized void removeAll() {
+    removeAllResult = agent.removeAllIdentities();
+  }
+
+  @Override
+  public synchronized boolean removeAllIdentities() {
+    removeAllResult = true;
+    // Preserve dispatch to removeAll() for subclasses compiled before this method was added.
+    removeAll();
+    return removeAllResult;
```
norrisjeremy, line 65: I would lean toward not adding a `removeAllResult` field, and instead simply do this:
```java
@Override
public void removeAll() {
  agent.removeAllIdentities()(;
}

@Override
public boolean removeAllIdentities() {
  return agent.removeAllIdentities();
}
```

src/main/java/com/jcraft/jsch/IdentityRepositoryWrapper.java
```diff
@@ -64,7 +64,25 @@ public boolean add(byte[] identity) {
...
+    if (blob == null)
+      return false;
+    boolean removed = false;
+    for (int i = cache.size() - 1; i >= 0; i--) {
+      Identity identity = cache.elementAt(i);
+      byte[] _blob = identity.getPublicKeyBlob();
+      if (_blob == null || !Util.array_equals(blob, _blob))
+        continue;
+      cache.removeElementAt(i);
+      identity.clear();
```
norrisjeremy: I'm not sure if we should be clearing an identity that is present in the cache.

src/main/java/com/jcraft/jsch/IdentityRepositoryWrapper.java
```diff
@@ -64,7 +64,25 @@ public boolean add(byte[] identity) {
 
   @Override
   public boolean remove(byte[] blob) {
-    return ir.remove(blob);
+    // getIdentities() hands out the cached identities as well, so they have to be removable too.
+    boolean removed = removeFromCache(blob);
+    return ir.remove(blob) || removed;
+  }
+
+  private synchronized boolean removeFromCache(byte[] blob) {
```
norrisjeremy, line 72: Nothing else in this class is synchronized (specifically, the `add()` method can insert entries into the cache while this method is running).
So I wonder if it would be best to just omit it (since everything else in this class seems to rely on single-threaded access to the `cache`).

src/main/java/com/jcraft/jsch/IdentityRepository.java
```diff
@@ -44,4 +44,21 @@ public interface IdentityRepository {
   public boolean remove(byte[] blob);
 
   public void removeAll();
```
norrisjeremy, line 46: I wonder if we should also mark this method as `@Deprecated` in order to encourage users to switch to using the new `removeAllIdentities()` method?

src/main/java/com/jcraft/jsch/IdentityRepositoryWrapper.java
norrisjeremy, line 82: The `removeAll()` and `removeAllIdentities()` methods don't clear Identity instances that are cleared from the cache, so I wonder if it should be omitted here as well?

norrisjeremy:
Hi @fornwall,

Just wanted to check and see if you had a chance to review my comments?

Thanks,
Jeremy

PR #1118, opened 2026-08-10
Add Session.getDisconnectCause()

fornwall:
Adds `Session.getDisconnectCause()` so callers can inspect why a session ended unexpectedly.

Exceptions that stop the read loop are retained after teardown. Local disconnects are ignored, `connect()` clears the previous cause, and the getter returns `null` while connected.


norrisjeremy:
I strongly disagree with adding this.

fornwall:
@norrisjeremy Thanks for the feedback - closing this one. Let me know if you think some alternative API could serve the same need and I can try that.

PR #1119, opened 2026-08-10
Expose verified host certificates through Session

fornwall:
## Summary

Expose the verified OpenSSH host certificate through `Session#getHostKeyCertificate()`.

Successful certificate verification currently returns from `Session#checkHost` before `doCheckHostKey`, leaving `Session#getHostKey()` unset and the certificate inaccessible to callers. This change:

- records the underlying public key in `Session#getHostKey()`;
- retains the verified certificate in `Session#getHostKeyCertificate()`;
- makes `OpenSshCertificate` and its read accessors public.

`getHostKeyCertificate()` remains `null` for plain host-key authentication and when certificate verification fails but `host_certificate_to_key_fallback` accepts the underlying key.

## Tests

`./mvnw -B -DskipITs=false -Dit.test=HostCertificateIT verify`

- 487 unit tests passed
- 11 `HostCertificateIT` tests passed
- formatting, import sorting, and forbidden-API checks passed

Disclaimar: Created with claude code and opus 5, reviewed by me.

### DIFF
--- a/src/main/java/com/jcraft/jsch/Session.java
+++ b/src/main/java/com/jcraft/jsch/Session.java
@@ -214,6 +214,9 @@ public class Session {
       throw new JSchException("session is already connected");
     }
     initialKex = true;
+    // Retain these values after disconnect, but not across connection attempts.
+    hostkey = null;
+    hostCertificate = null;
 
     io = new IO();
     if (random == null) {
@@ -972,6 +975,9 @@ public class Session {
           getLogger().log(Logger.INFO, "Host '" + chost + "' is known and matches the "
               + kex.getKeyType() + " host certificate");
         }
+        // This path does not reach doCheckHostKey, which normally records the host key.
+        hostkey = new HostKey(chost, certificate.getCertificatePublicKey());
+        hostCertificate = certificate;
         return;
       } catch (JSchRevokedHostKeyException e) {
         // Revoked key is a hard failure — never fall back to plain key checking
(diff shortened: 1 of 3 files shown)


norrisjeremy:
What's the purpose of this?
We don't currently return the verified host key, so I'm not sure I understand the reason to return the verified host certificate?

norrisjeremy:
> What's the purpose of this? We don't currently return the verified host key, so I'm not sure I understand the reason to return the verified host certificate?

Nvm, let me think on this some more.
I'm not sure I'm comfortable with the approach taken here.

src/main/java/com/jcraft/jsch/Session.java
```diff
@@ -3289,10 +3295,28 @@ public void noMoreSessionChannels() throws Exception {
...
+   * <p>
+   * This is {@code null} when the server used a plain host key or when certificate verification
+   * failed and {@code host_certificate_to_key_fallback} accepted its plain key instead. The value
+   * is retained after disconnect and cleared when a new connection attempt begins.
+   * </p>
+   *
+   * @return the verified host certificate, or {@code null} if none was used
+   * @see #getHostKey()
+   */
+  public OpenSshCertificate getHostKeyCertificate() {
```
norrisjeremy, line 3316: I am opposed to making the `OpenSshCertificate` class a part of the public API.

fornwall: 👍 Thanks for the feedback! Let me know if you think the certificate information can be exposed in some other way and I'll try that out. Otherwise I'll close this PR as a wont do.

norrisjeremy: I need to think through this some more to try and think of a solution that wouldn't leak the internal `OpenSshCertificate` class.

norrisjeremy:
FYI, I haven't forgotten about this and am working on an alternative solution.

norrisjeremy:
Hi @fornwall,

Just wanted to update you that I haven't forgotten about this.
I'm currently waiting on a clarification from the author of draft-ietf-sshm-cert on how a few fields are structured.

Thanks,
Jeremy

norrisjeremy:
I still haven't forgotten about this, but I'm still waiting to hear back clarification from the author of draft-ietf-sshm-cert on how a few fields are structured.

Issue #1115, opened 2026-08-10
Support ML-KEM with Java 21 and 17 (from 10/2026 patch day) and Java 11 and 8 (H2 2027)


beth-soptim:
Starting with October patch day Java 21 and 17 will support ML-KEM:

https://www.java.com/de/jre-jdk-cryptoroadmap.html

OpenJDK is also following this roadmap:

https://bugs.openjdk.org/browse/JDK-8298390

21 -> https://bugs.openjdk.org/browse/JDK-8389458
17 -> https://bugs.openjdk.org/browse/JDK-8389616

The backports to 11 and 8 are planned for H2 2027.

Please note that Java 17 also includes KEM api since version 17.0.13 (Java SE 17 Maintenance Release 1 -> https://jcp.org/aboutJava/communityprocess/maintenance/jsr392/jsr392-mr1-changes.html)

mwiede:
Thanks for this info. Once the JDKs are published, the ML-KEM classes can be compiled with 17+21 and the multi-release jar can be reorganized.

beth-soptim:
I think this needs some special treatment because Java 25 compiler seems to refuse compilation because 25 does not include metadata that KEM API was also added in SE 17 MR 1.

Maven toolchains might be required (= using a JDK 17 to compile Java 17 classes).

norrisjeremy:
Hi @beth-soptim,

I suspect that the reason #1120 fails to compile is because we need to wait for the next Java 25 release in October, which hopefully will be updated to include the SE 17 MR 1 changes when compiling for release=17.

Thanks,
Jeremy

beth-soptim:
Hi,

I think it will be required to use 17 to compile 17 source code with KEM API because 21, 25, ... contain the "since 21" info in ct.sym file.

The backport of the KEM API has already happenened in 2024 in version 17.0.13 -> https://bugs.openjdk.org/browse/JDK-8336893

It seems that only in 17 the ct.sym file contains the "correct" information.

https://github.com/openjdk/jdk17u-dev/blame/master/src/java.base/share/classes/javax/crypto/KEM.java (since 21)
vs.
https://github.com/openjdk/jdk21u-dev/blame/master/src/java.base/share/classes/javax/crypto/KEM.java (since 17, with "This class is defined in Java SE 17 Maintenance Release 1")

norrisjeremy:
Hi @beth-soptim,

[...]

PR #1120, opened 2026-08-12
Support ML-KEM with Java 21 and 17 (from 10/2026 patch day)

beth-soptim:
see #1115


norrisjeremy (changes requested):
We will need to wait to merge this until after the October 2026 releases are dropped.

pom.xml
```diff
@@ -389,16 +389,31 @@
...
+                        <configuration>
+                            <release>8</release>
+                            <showDeprecation>true</showDeprecation>
+                            <showWarnings>true</showWarnings>
+                            <failOnWarning>true</failOnWarning>
+                            <compilerArgs>
+                                <arg>-Xlint:all,-processing,-classfile,-options</arg>
+                            </compilerArgs>
+                        </configuration>
+                    </execution>
```
norrisjeremy, line 416: What is the purpose of these changes?

pom.xml
```diff
@@ -464,6 +479,27 @@
...
+                                <compileSourceRoot>${project.basedir}/src/main/java17</compileSourceRoot>
+                            </compileSourceRoots>
+                            <!-- not via <release> because of backported KEM API in SE 17 MR 1 -->
+                            <outputDirectory>${project.build.outputDirectory}/META-INF/versions/17</outputDirectory>
+                            <source>17</source>
+                            <target>17</target>
+                            <showWarnings>false</showWarnings>
+                            <failOnWarning>false</failOnWarning>
+                            <compilerArgs>
+                                <arg>-XDignore.symbol.file</arg>
```
norrisjeremy: Is this compiler argument documented somewhere?
All the references I find indicate that it is an undocumented internal argument, and could be removed or changed at some future point.

norrisjeremy (changes requested):
These changes need to wait until after the October releases are published.
Then we can reevaluate whether to move forward with them or not.

pom.xml
beth-soptim: I have removed the usage of -XDignore.symbol.file.

pom.xml
beth-soptim: default-compile and default-testCompile (which are the default IDs) are needed because for 17 source/target instead of release must be used to avoid usage of symbol file.

norrisjeremy: That doesn't make any sense to me?
The configuration here is only for sources under `src/main/java` + `src/test/java`.
Why would that have any impact on the configuration for Java 17 below (which is for `src/main/java17`)?

norrisjeremy: Ok, I think I see why it's needed now.

norrisjeremy (commented):
I'm wondering if we would be better inclined to simply create a different artifact that contains Java 17 versions of ML-KEM implementations instead of trying to coerce it directly into JSch itself. Then users could optionally add it as dependency and configure JSch to use it if so desired.

Issue #1127, opened 2026-08-15
SftpATTRS.getATTR allocates an array sized from an unbounded server-supplied count


brimmasch:
In `SftpATTRS.getATTR` the extended-attribute count comes straight off the wire and sizes the array, with only a lower bound between the two (2.28.6, lines 190-193):

```java
if ((flags & SSH_FILEXFER_ATTR_EXTENDED) != 0) {
    int count = buf.getInt();
    if (count > 0) {
        attr.extended = new String[count * 2];
        ...
    }
}
```

A hostile or MITM'd server sends a large count in a handful of bytes and the client allocates immediately - before reading any attribute data. Large values make `count * 2` overflow negative and throw `NegativeArraySizeException`; just under that it is a multi-gigabyte allocation and an `OutOfMemoryError`. Either one kills the transfer, and an OOM inside a shared application server is not reliably contained to the thread that caused it.

Reachable from anything that parses attributes, so `ls` and `get` included.

The amplification is what makes it worth bounding - roughly four bytes of response for a 2^31 allocation attempt.

Sanity-checking `count` against what is actually left in the buffer would be enough. Each extended attribute is two length-prefixed strings, so eight bytes minimum per attribute even when both are empty:

```java
int count = buf.getInt();
if (count > 0) {
    // two length-prefixed strings per attribute, so 8 bytes minimum each
    if (count > buf.getLength() / 8) {
        throw new IllegalArgumentException("extended attribute count out of range: " + count);
    }
    attr.extended = new String[count * 2];
    ...
}
```

`getATTR` has no `throws` clause, so that is unchecked as written - clamping would work as well if you would rather not change the failure mode.

norrisjeremy:
Hi @brimmasch,

How could a server be MitM'd at the point in which this call occurs?
Also, if this was a hostile server, aren't there numerous other ways they could harm the client besides just this?

Can you cite any other SSH implementations that apply restrictions similar to what you are proposing?

Thanks,
Jeremy

brimmasch:
Good point.  Drop MitM and just look at it from a hostile server perspective.  I know it's a bit of a weird situation to be connecting to an unknown, hostile server but it could happen (and this is dinging me on a security audit).

The thing I actually care about is the threat of a hostile server being able to OOM me.

I agree there are other ways to harm the client but this one is real depending on how the jsch jar is loaded and consumed.  Being able to allocate GBs of memory is probably something worth preventing.

[...]

PR #1128, opened 2026-08-17
2.28.7 changes

norrisjeremy:
- #1127: add bounds checks when parsing SFTP vendor-specific extensions.
- Treat the reserved field in certs as an opaque byte array instead of a String.

### DIFF
--- a/src/main/java/com/jcraft/jsch/OpenSshCertificate.java
+++ b/src/main/java/com/jcraft/jsch/OpenSshCertificate.java
@@ -102,7 +102,7 @@ class OpenSshCertificate {
   /**
    * Reserved field for future use
    */
-  private final String reserved;
+  private final byte[] reserved;
 
   /**
    * The CA's key that signed this certificate
@@ -184,8 +184,8 @@ class OpenSshCertificate {
     return extensions == null ? null : Collections.unmodifiableMap(extensions);
   }
 
-  String getReserved() {
-    return reserved;
+  byte[] getReserved() {
+    return reserved == null ? null : reserved.clone();
   }
 
   byte[] getSignatureKey() {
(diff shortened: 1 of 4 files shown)


norrisjeremy:
Hi @luigidemasi,

Can you review the changes I made to the reserved field in certs as an opaque byte array instead of a String?
I don't think that parsing it as a Java String type is correct, as it isn't defined as a character string in [draft-ietf-sshm-cert-01](https://datatracker.ietf.org/doc/html/draft-ietf-sshm-cert-01.html), but rather is simply an opaque field that could be used in the future to contain any type of data (that may not be valid UTF-8).
```
The "reserved" field is reserved for future use. Implementations MUST ignore the contents of this field if it is not empty.
```

Thanks!
Jeremy

luigidemasi:
Hi @norrisjeremy,

I reviewed the change and I agree with treating `reserved` as an opaque byte array. The field is declared as SSH wire `string`, and per [RFC 4251](https://datatracker.ietf.org/doc/html/rfc4251#section-5) that is arbitrary binary data.

UTF-8 decoding is only appropriate where the spec assigns text semantics, like key id and valid principals. 

On top of that, decoding with UTF-8 is lossy: malformed sequences are replaced with U+FFFD, so if a future revision ever puts binary data in reserved, the String representation would corrupt it silently.

I also checked the implementation side:

- getReserved() has no callers in main or test code, so there is no behavioral impact.

- Signature verification is unaffected, since it uses the raw message bytes captured directly from the buffer.

- The clone() on get matches the pattern of the other byte[] getters (nonce, signatureKey, signature, message).

Looks good to me.

norrisjeremy:
Hi @luigidemasi,

Great, thanks for the quick review!

Thanks,
Jeremy

PR #1145, opened 2026-09-07
Expose local SFTP window/packet size setters as public

courville:
ChannelSftp caps every pipelined SFTP READ request at a fixed 32KB local packet size (LOCAL_MAXIMUM_PACKET_SIZE), which limits transfer throughput even with multiple requests in flight. Channel already had setLocalWindowSizeMax/setLocalWindowSize/setLocalPacketSize setters that could raise these values per-channel before connecting, but they were package-private, so callers had no way to tune them.

Making them public lets consumers configure a larger packet/window size from their own code without having to fork the library default.

This enable to catch up throughput penalty and beat sshj implementation in my application nova video player cf. https://github.com/nova-video-player/aos-AVP/issues/1943

### DIFF
--- a/src/main/java/com/jcraft/jsch/Channel.java
+++ b/src/main/java/com/jcraft/jsch/Channel.java
@@ -46,9 +46,13 @@ public abstract class Channel {
 
   private static final AtomicInteger index = new AtomicInteger();
 
-  // sanity cap for setLocalPacketSize: lmpsize is used to allocate a Buffer of that size
-  // (e.g. ChannelSftp.start()), so an unbounded value could trigger an excessive allocation
-  private static final int MAX_LOCAL_PACKET_SIZE = 16 * 1024 * 1024;
+  // sanity cap for setLocalPacketSize: lmpsize is both used to allocate a Buffer of that size
+  // (e.g. ChannelSftp.start()) and advertised to the server as the max size of a single
+  // CHANNEL_DATA payload it may send us. That payload is wrapped in an SSH packet together with
+  // ~9 bytes of channel-data framing plus padding, so it must stay comfortably under
+  // Session.PACKET_MAX_SIZE (RFC 4253 6.1 Maximum Packet Length) or the resulting packet gets
+  // discarded by the transport layer, killing the connection.
+  private static final int MAX_LOCAL_PACKET_SIZE = Session.PACKET_MAX_SIZE - 4096;
 
   int id;
   volatile int recipient = -1;
--- a/src/main/java/com/jcraft/jsch/Channel.java
+++ b/src/main/java/com/jcraft/jsch/Channel.java
@@ -406,16 +406,35 @@ public abstract class Channel {
     }
   }
 
-  public void setLocalWindowSizeMax(int foo) {
-    this.lwsize_max = foo;
+  public void setLocalWindowSizeMax(int size) {
+    if (isConnected()) {
+      throw new IllegalStateException("local window size max cannot be changed after channel is connected");
+    }
+    if (size <= 0) {
+      throw new IllegalArgumentException("local window size max must be positive: " + size);
+    }
+    this.lwsize_max = size;
   }
 
-  public void setLocalWindowSize(int foo) {
-    this.lwsize = foo;
+  public void setLocalWindowSize(int size) {
+    if (isConnected()) {
+      throw new IllegalStateException("local window size cannot be changed after channel is connected");
+    }
+    if (size <= 0 || size > lwsize_max) {
+      throw new IllegalArgumentException(
+          "local window size must be positive and not exceed local window size max (" + lwsize_max + "): " + size);
+    }
+    this.lwsize = size;
   }
 
-  public void setLocalPacketSize(int foo) {
-    this.lmpsize = foo;
+  public void setLocalPacketSize(int size) {
+    if (isConnected()) {
+      throw new IllegalStateException("local packet size cannot be changed after channel is connected");
+    }
+    if (size <= 0) {
+      throw new IllegalArgumentException("local packet size must be positive: " + size);
+    }
+    this.lmpsize = size;
   }
 
   synchronized void setRemoteWindowSize(long foo) {
--- a/src/main/java/com/jcraft/jsch/Channel.java
+++ b/src/main/java/com/jcraft/jsch/Channel.java
@@ -46,6 +46,10 @@ public abstract class Channel {
 
   private static final AtomicInteger index = new AtomicInteger();
 
+  // sanity cap for setLocalPacketSize: lmpsize is used to allocate a Buffer of that size
+  // (e.g. ChannelSftp.start()), so an unbounded value could trigger an excessive allocation
+  private static final int MAX_LOCAL_PACKET_SIZE = 16 * 1024 * 1024;
+
   int id;
   volatile int recipient = -1;
   protected byte[] type = Util.str2byte("foo");
@@ -431,8 +435,9 @@ public abstract class Channel {
     if (isConnected()) {
       throw new IllegalStateException("local packet size cannot be changed after channel is connected");
     }
-    if (size <= 0) {
-      throw new IllegalArgumentException("local packet size must be positive: " + size);
+    if (size <= 0 || size > MAX_LOCAL_PACKET_SIZE) {
+      throw new IllegalArgumentException(
+          "local packet size must be positive and not exceed " + MAX_LOCAL_PACKET_SIZE + ": " + size);
     }
     this.lmpsize = size;
   }
--- a/src/main/java/com/jcraft/jsch/Channel.java
+++ b/src/main/java/com/jcraft/jsch/Channel.java
@@ -406,15 +406,15 @@ public abstract class Channel {
     }
   }
 
-  void setLocalWindowSizeMax(int foo) {
+  public void setLocalWindowSizeMax(int foo) {
     this.lwsize_max = foo;
   }
 
-  void setLocalWindowSize(int foo) {
+  public void setLocalWindowSize(int foo) {
     this.lwsize = foo;
   }
 
-  void setLocalPacketSize(int foo) {
+  public void setLocalPacketSize(int foo) {
     this.lmpsize = foo;
   }
 
--- a/src/main/java/com/jcraft/jsch/Channel.java
+++ b/src/main/java/com/jcraft/jsch/Channel.java
@@ -438,6 +438,21 @@ public abstract class Channel {
     this.lwsize = size;
   }
 
+  // Session's own flow-control bookkeeping (see SSH_MSG_CHANNEL_DATA/SSH_MSG_CHANNEL_EXTENDED_DATA
+  // handling) needs to update the local window continuously while the channel is connected, and can
+  // legitimately drive it down to 0 before resetting it to lwsize_max. That is a different contract
+  // than the public setLocalWindowSize, which is meant for pre-connect configuration by consumers,
+  // so keep this bookkeeping update package-private and unrestricted (besides the lwsize_max
+  // bound).
+  void updateLocalWindowSize(int size) {
+    if (size < 0 || size > lwsize_max) {
+      throw new IllegalArgumentException(
+          "local window size must not be negative and must not exceed local window size max ("
+              + lwsize_max + "): " + size);
+    }
+    this.lwsize = size;
+  }
+
   public void setLocalPacketSize(int size) {
     if (isConnected()) {
       throw new IllegalStateException(
--- a/src/main/java/com/jcraft/jsch/Channel.java
+++ b/src/main/java/com/jcraft/jsch/Channel.java
@@ -416,7 +416,8 @@ public abstract class Channel {
 
   public void setLocalWindowSizeMax(int size) {
     if (isConnected()) {
-      throw new IllegalStateException("local window size max cannot be changed after channel is connected");
+      throw new IllegalStateException(
+          "local window size max cannot be changed after channel is connected");
     }
     if (size <= 0) {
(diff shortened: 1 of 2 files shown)


norrisjeremy:
Hi @courville,

These changes appear to cause numerous test failures.
Before we could even consider merging the changes, the reason for the test failures would need to be identified and resolved.

Thanks,
Jeremy

courville:
Yes... fighting two conflicting review requirements. I will investigate but this is a pain for such a simple addition.

src/main/java/com/jcraft/jsch/Channel.java
```diff
@@ -406,16 +414,55 @@ synchronized void checkSpace(int len) throws IOException {
...
-    this.lwsize_max = foo;
+  public void setLocalWindowSizeMax(int size) {
+    if (isConnected()) {
+      throw new IllegalStateException(
+          "local window size max cannot be changed after channel is connected");
+    }
+    if (size <= 0) {
+      throw new IllegalArgumentException("local window size max must be positive: " + size);
+    }
+    this.lwsize_max = size;
```
norrisjeremy, line 425: It makes me a bit nervous to allow users to set lwsize_max to an unbounded upper value.
Do you know if other SSH implementations allow an unbound upper value or do they enforce a cap?

norrisjeremy (commented):
Instead of throwing an unchecked `IllegalArgumentException`, could we make these methods instead throw a checked `JSchException`?
Or would throwing a checked exception force other changes to places that JSch internally calls these methods?

src/main/java/com/jcraft/jsch/Channel.java
courville: sshj (net.schmizz.sshj) does not enforce an upper bound on window size.

- ConnectionImpl.setWindowSize(long windowSize): plain setter, no validation at all (not even a positive-value check). Default is 2048 * 1024 (2MB).
- ConnectionImpl.setMaxPacketSize(int maxPacketSize): only checks maxPacketSize > 0; no upper cap. Default is 32 * 1024.
- Window / Window.Local / Window.Remote constructors: only validate maxPacketSize > 0. The window size argument itself is not validated.

Source: hierynomus/sshj, src/main/java/net/schmizz/sshj/connection/ConnectionImpl.java and src/main/java/net/schmizz/sshj/connection/channel/Window.java.

norrisjeremy: Thank you for the information about sshj.
Do you know how implementations handle this, such as OpenSSH?

norrisjeremy:
> Instead of throwing an unchecked `IllegalArgumentException`, could we make these methods instead throw a checked `JSchException`? Or would throwing a checked exception force other changes to places that JSch internally calls these methods?

Hi @courville,

Did you have an opportunity to review the question above?

Thanks,
Jeremy

courville:
To my current understanding, the answer depends on which point in the PR you compare against.

At the base of this PR, `setLocalWindowSize` was already called internally by `Session.java` (4 call sites, inside the packet-handling loop in `run()`), package-private and unvalidated. Switching it to a checked `JSchException` at that point would have forced changes to those internal call sites, since `run()` does not declare any checked exception and each call would need its own try/catch.

That dependency has since been removed within this PR: `Session.java`'s internal call sites now use a separate package-private `updateLocalWindowSize` method instead, decoupling the internal flow-control bookkeeping from the public setters. As things stand now, no code in this repository calls `setLocalWindowSizeMax`, `setLocalWindowSize`, or `setLocalPacketSize` other than their own definitions in `Channel.java`, so switching these three to a checked `JSchException` would not require any further internal changes.

src/main/java/com/jcraft/jsch/Channel.java
```diff
@@ -46,6 +46,14 @@ public abstract class Channel {
 
   private static final AtomicInteger index = new AtomicInteger();
 
+  // sanity cap for setLocalPacketSize: lmpsize is both used to allocate a Buffer of that size
+  // (e.g. ChannelSftp.start()) and advertised to the server as the max size of a single
+  // CHANNEL_DATA payload it may send us. That payload is wrapped in an SSH packet together with
+  // ~9 bytes of channel-data framing plus padding, so it must stay comfortably under
+  // Session.PACKET_MAX_SIZE (RFC 4253 6.1 Maximum Packet Length) or the resulting packet gets
+  // discarded by the transport layer, killing the connection.
+  private static final int MAX_LOCAL_PACKET_SIZE = Session.PACKET_MAX_SIZE - 4096;
```
norrisjeremy, line 55: Do we know what max values other implementations use?

courville: To my current understanding, neither sshj nor OpenSSH caps local max packet size beyond the RFC 4253 transport packet-length ceiling: sshj's `setMaxPacketSize` only rejects values `<= 0` (default 32KB, no upper bound), and OpenSSH hardcodes `CHAN_SES_PACKET_DEFAULT` at 32KB with no user-configurable option, while its own `PACKET_MAX_SIZE` transport ceiling is 256KB, the same value and name jsch uses.

norrisjeremy: Is it necessary to expose setting both the max packet size and max window size in order to obtain the performance increase you desire? Or is exposing just one or the other sufficient?

courville: Only `setLocalPacketSize` is strictly necessary: with the default 2MB window and 16 max pipelined requests, in-flight data stays well under the window even at 64KB packets, so the window increase shouldn't be required by that math. That said, we only benchmarked both changes together, not packet size alone, so this isn't empirically confirmed. However IMHO symmetry of setters should be present for parity's sake.

src/main/java/com/jcraft/jsch/Channel.java
```diff
@@ -406,16 +414,55 @@ synchronized void checkSpace(int len) throws IOException {
...
-  void setLocalWindowSize(int foo) {
-    this.lwsize = foo;
+  // Session's own flow-control bookkeeping (see SSH_MSG_CHANNEL_DATA/SSH_MSG_CHANNEL_EXTENDED_DATA
+  // handling) needs to update the local window continuously while the channel is connected, and can
+  // legitimately drive it down to 0 before resetting it to lwsize_max. That is a different contract
+  // than the public setLocalWindowSize, which is meant for pre-connect configuration by consumers,
+  // so keep this bookkeeping update package-private and unrestricted (besides the lwsize_max
+  // bound).
+  void updateLocalWindowSize(int size) {
+    if (size < 0 || size > lwsize_max) {
```
norrisjeremy: I'm mildly concerned about enforcing a non-negative size here. as JSch hasn't historically enforced that for the previous `setLocalWindowSize()` uses.
It appears JSch previously would have allowed a server to burst beyond the current lwsize (all the current uses simply call `setLocalWindowSize(channel.lwsize - len)`, so if `len` exceeds `lwsize`, it could go negative).

Since the purpose of `updateLocalWindowSize()` is to for internal flow-control bookkeeping (as you mention in the comment), perhaps it would be best to simply not enforce any particular constraints on the value (instead trusting internal usages to do the right thing)?
I.e., just make `updateLocalWindowSize()` directly set the `this.lwsize = foo;` regardless of what value `foo` is?

courville: All the changes to the PR were addressing the comments from sonarqube raised during the CI ... I think the original PR was much simpler.

norrisjeremy: I don't especially care if SonarQube raised it as an issue: we can't account for how random servers in the wild behave and I'm not willing to risk JSch suddenly closing connections because a random server sent a packet that exceeded the current window size.
Please change the `updateLocalWindowSize()` function exactly how `setLocalWindowsSize()` previously functioned (by just setting `this.lwsize = foo;` so that we don't have to worry about causing a regression for users.

courville: Done. `updateLocalWindowSize()` has been simplified to directly assign `this.lwsize = size;` without any bounds checking, matching the exact prior behavior of `setLocalWindowSize()` in `Session.run()`.

src/main/java/com/jcraft/jsch/Channel.java
courville: Regarding checked JSchException vs IllegalArgumentException/IllegalStateException: since
Session.run() now uses updateLocalWindowSize(), switching the public setters to throws JSchException
would not touch any internal call sites. I am happy to switch them to JSchException if you prefer
that for JSch's API conventions, or leave them as unchecked argument/state exceptions.

norrisjeremy: Let's go ahead and switch them to throwing checked JSchException's so that users are aware of the potential for an exception being thrown.
Can you also add Javadocs to both these methods, being sure to highlight the reasons why a JSchException could be thrown too?

src/main/java/com/jcraft/jsch/Channel.java
norrisjeremy: Thx!

courville:
If need be I can squash all back to simpler b28b683 (which was my original proposal and probably should have sticked to it...).

norrisjeremy:
> If need be I can squash all back to simpler [b28b683](https://github.com/mwiede/jsch/commit/b28b683d83457fb99bbd7fad81d619523a051acb) (which was my original proposal and probably should have sticked to it...).

No, I do agree that adding bounds checks to the publicly exposed versions of these methods as you've done is prudent, and switching the internal call sites to a non-bounds check version maintains compatibility without worry for introducing any possible regressions.

I think the only remaining changes are switching the public versions over to throwing checked JSchExceptions and adding Javadocs to them (in particular highlighting the conditions in which a JSchException could be thrown), this way users understand the API contract.

norrisjeremy:
I wonder if we should add public getters in addition to publicly exposing the setters, so that users can successfully determine what the packet & window size values are?

src/main/java/com/jcraft/jsch/Channel.java
```diff
@@ -406,16 +414,50 @@ synchronized void checkSpace(int len) throws IOException {
...
   }
 
-  void setLocalWindowSizeMax(int foo) {
-    this.lwsize_max = foo;
+  public void setLocalWindowSizeMax(int size) {
+    if (isConnected()) {
+      throw new IllegalStateException(
+          "local window size max cannot be changed after channel is connected");
+    }
+    if (size <= 0) {
```
norrisjeremy, line 428: I wonder if we should disallow setting lwsize_max to < lwsize?

norrisjeremy: In fact, I think we definitely should enforce that lwsize_max >= lwsize.
Otherwise, in Session.java when we send a window update, we could end up sending an absurd value to the server (`buf.putInt(channel.lwsize_max - channel.lwsize);`).

courville:
All requested adjustments have been made:

1. Switched `setLocalWindowSizeMax`, `setLocalWindowSize`, and `setLocalPacketSize` to throw checked `JSchException`.
2. Enforced `size >= lwsize` in `setLocalWindowSizeMax` to avoid sending negative/underflow window adjust deltas to the server.
3. Added public getters: `getLocalWindowSizeMax()`, `getLocalWindowSize()`, and `getLocalPacketSize()`.
4. Added Javadoc documentation to all new getters and setters detailing arguments and exact conditions triggering `JSchException`.

norrisjeremy:
Hi @courville,

I've been giving this more thought and I think that allowing users to independently set lwsize + lwsize_max doesn't really make sense, because I can't think of a use case in which you want these values to be different when a new Channel starts.

I'm wondering if a better approach would be to only expose `setLocalWindowSizeMax()` and keep `setLocalWindowSize()` as an internal method that is just used for the internal bookkeeping like it is today.
And instead change `setLocalWindowSizeMax()` to set both `lwsize` & `lwsize_max` to the user supplied size value (with the constraints that the Channnel isn't connected and that the provided size value isn't `<= 0`).

What do you think?

Thanks,
Jeremy

courville:
Hi Jeremy,

I completely agree — that is a much cleaner design and avoids unnecessary configuration complexity.

I've updated the PR accordingly:
- `setLocalWindowSizeMax(int)` is public, validates `!isConnected()` and `size > 0` (throwing checked `JSchException`), and sets both `lwsize_max` and `lwsize` to the specified size.
- `setLocalWindowSize(int)` remains an unconstrained, package-private method for `Session.run()`'s internal flow-control bookkeeping (which also lets us remove `updateLocalWindowSize`).
- Kept the public getters (`getLocalWindowSizeMax()`, `getLocalWindowSize()`, `getLocalPacketSize()`) and the public `setLocalPacketSize(int)` with their Javadoc documentation.

Thanks!

norrisjeremy:
Hi @courville,

Great thanks!
I think we can omit the `getLocalWindowSize()` method, as I don't think it offers much benefit for users.
Functionally, I think all that user should care about is `getLocalWindowSizeMax()` (i.e. `lwsize_max`).
Do you agree?

Thanks!
Jeremy

courville:
Hi Jeremy,

I agree completely — since `lwsize` fluctuates dynamically with incoming packets, exposing it doesn't provide real value, and `getLocalWindowSizeMax()` already represents the configured window capacity.

I've removed `getLocalWindowSize()`. The public API is now clean and symmetrical:
- `setLocalWindowSizeMax(int)` / `getLocalWindowSizeMax()`
- `setLocalPacketSize(int)` / `getLocalPacketSize()`

Thanks!

norrisjeremy (approved):
Hi @courville,

Thank you for your patience working on this.
I think this looks good now.

Thanks!
Jeremy

PR #1159, opened 2026-09-20
Use the platform JCA for EdDSA on Android

FireBurn:
## Problem

On Android, `ssh-ed25519` public key authentication cannot succeed:

```
com.jcraft.jsch.JSchException: Auth cancel for methods 'publickey,keyboard-interactive'
Signature algorithms unavailable for non-agent identities = [ssh-ed25519, ssh-ed448]
```

`JSch` picks its EdDSA implementation from `JavaVersion.getVersion()`. Android always reports 8 and
ignores `META-INF/versions`, so `ssh-ed25519` is configured to the BouncyCastle implementation.
BouncyCastle is an optional dependency and is normally absent there, so `Session.checkSignatures()`
drops the algorithm. Android's JCA has provided Ed25519 since API 28.

## Change

- The Java 8 `jce.SignatureEdDSA` gets a working implementation in place of the throwing stubs,
  using `Signature`/`KeyFactory` with the `"Ed25519"`/`"Ed448"` algorithm names.
- It is selected only on Android. Java 15 and later keep the `META-INF/versions/15` classes, Java 8
  to 14 keeps BouncyCastle, and nothing else moves.
- `keypairgen.eddsa` stays version based: key generation through JSch on Android still needs
  BouncyCastle.

### Why the keys are wrapped in DER

The Java 15 implementation builds keys from `EdECPublicKeySpec` and `EdECPrivateKeySpec`, which do
not exist at Java 8 source level. This class has to compile at `--release 8` and is the only one
Android ever loads, so the raw SSH key bytes are wrapped in their RFC 8410 `SubjectPublicKeyInfo`
and `PrivateKeyInfo` encodings and passed as `X509EncodedKeySpec`/`PKCS8EncodedKeySpec`. Android
does expose the `EdEC` specs from API 33, but they cannot be referenced from a class that must also
compile for Java 8.

### No probing in a static initializer

The earlier attempt at this (#1159) decided by constructing and initializing a `Signature` inside
`JSch`'s static initializer. That is gone. The runtime is identified from `java.vm.name` and
`java.vendor`, and nothing is constructed at class initialization time.

## Behaviour on other runtimes

Unchanged. Measured on a JDK 8 JVM (1.8.0_504) against 2.28.7 and against this branch:

| config key | 2.28.7 | this branch |
|---|---|---|
| `ssh-ed25519` | `bc.SignatureEd25519` | `bc.SignatureEd25519` |
| `ssh-ed448` | `bc.SignatureEd448` | `bc.SignatureEd448` |
| `keypairgen.eddsa` | `bc.KeyPairGenEdDSA` | `bc.KeyPairGenEdDSA` |

The one difference is for anyone who deliberately configures the `jce` class on a runtime without
EdDSA: the constructor used to throw `UnsupportedOperationException`, and now `init()` throws
`NoSuchAlgorithmException`. `Session.checkSignatures()` catches `Exception | LinkageError`, so the
algorithm is marked unavailable either way.

## Tests

- `com.jcraft.jsch.jce.SignatureEdDSATest` signs with this implementation and verifies with the
  platform's own `Signature`, and the reverse, for Ed25519 and Ed448, plus the short key guard.
  Surefire runs against `target/classes`, where the `META-INF/versions` entries are inert, so these
  exercise the Java 8 classes, which are the ones Android loads. Gated to Java 15 and above,
  because the platform needs EdDSA in its JCA; the 8 and 11 matrix legs skip them.
- `JSchTest` pins the implementation chosen for a standard runtime and for an Android one.
- Full suite green on JDK 25 and on the JDK 8 leg.

## Checked on a device

OnePlus, Android 15 (API 35): `ssh-ed25519` publickey authentication and an interactive session
against OpenSSH 10.5. `ssh-ed448` is reported unavailable, since Android's JCA has no Ed448, which
is the same outcome as before this change.

## Note

`verify()` is the same unwrapping as the `bc` and Java 15 implementations. I left it alongside them
rather than extracting a shared helper, to keep the change contained. Happy to factor it out if you
would prefer that.


FireBurn:
Closing pending on-device verification; will reopen once confirmed.

src/main/java/com/jcraft/jsch/JSch.java
```diff
@@ -670,6 +674,29 @@ public static String getConfig(String key) {
...
+   */
+  private static void putIfAvailable(String key, String preferred, String fallback) {
+    if (canInstantiate(preferred)) {
+      config.put(key, preferred);
+    } else {
+      config.put(key, fallback);
+    }
+  }
+
+  private static boolean canInstantiate(String className) {
```
norrisjeremy, line 689: I strongly oppose coercing ed25519 support like this, by trying to initialize a signature implementation inside the class's static initializers.

src/main/java/com/jcraft/jsch/jce/SignatureEdDSA.java
```diff
@@ -16,49 +16,126 @@
...
+      tmp = new byte[j];
+      System.arraycopy(sig, i, tmp, 0, j);
+      sig = tmp;
+    }
+
+    return signature.verify(sig);
+  }
+
+  // RFC 8410 DER wrappers: SPKI for raw public keys, PKCS#8 for raw private seeds.
+  private static byte[] spkiPrefix(String algo) {
```
norrisjeremy, line 113: I don't understand the purpose of this code: is this needed because of some sort of implementation detail of how Android implements EdDSA?
We haven't needed to use it for EdDSA implementations present in Java 15+.

norrisjeremy (commented):
I have concerns about the way this is being coerced in here.

Is there a simple way to detect at runtime whether the underlying runtime is Android vs standard Java (so that the auto-configuration of this could be restricted to only executing on Android runtimes and not standard Java runtimes)?

PR #1160, opened 2026-09-20
Use the platform JCA for EdDSA on Android

FireBurn:
## Problem

On Android, `ssh-ed25519` public key authentication cannot succeed:

```
com.jcraft.jsch.JSchException: Auth cancel for methods 'publickey,keyboard-interactive'
Signature algorithms unavailable for non-agent identities = [ssh-ed25519, ssh-ed448]
```

Android ignores `META-INF/versions`, so JSch loads the base `JavaVersion`, which reports 8, and
`ssh-ed25519` is configured to BouncyCastle. BouncyCastle is an optional dependency and is normally
absent there, so `Session.checkSignatures()` drops the algorithm.

## Change

- The Java 8 `jce.SignatureEdDSA` gets a working implementation in place of the throwing stubs,
  using `Signature`/`KeyFactory` with the `"Ed25519"`/`"Ed448"` algorithm names.
- It is selected on Android runtimes only. Java 15 and later keep the `META-INF/versions/15`
  classes and Java 8 to 14 keeps BouncyCastle, both unchanged.
- `keypairgen.eddsa` stays as it was: key generation through JSch on Android still needs
  BouncyCastle.

There is no API level check. What the platform does not provide fails in `init()` and is dropped by
the availability check, which is what already happens when BouncyCastle is missing. That keeps the
change honest across releases rather than encoding a level that turns out to be wrong.

### Why the keys are wrapped in DER

The Java 15 implementation builds keys from `EdECPublicKeySpec` and `EdECPrivateKeySpec`, which do
not exist at Java 8 source level. This class has to compile at `--release 8` and is the only one
Android ever loads, so the raw SSH key bytes are wrapped in their RFC 8410 `SubjectPublicKeyInfo`
and `PrivateKeyInfo` encodings and passed as `X509EncodedKeySpec`/`PKCS8EncodedKeySpec`. Android
does expose the `EdEC` specs from API 33, but they cannot be referenced from a class that must also
compile for Java 8.

### Why both services come from one provider

`KeyFactory.getInstance("Ed25519")` is ambiguous on Android: the key store provider registers one
too, and it rejects imported key material. Taking the `Signature` and the `KeyFactory` from a
single provider avoids that. Where no provider offers both, the platform cannot sign with an
imported key at all, so the algorithm is left unavailable.

### No probing in a static initializer

The earlier attempt (#1159) decided by constructing and initializing a `Signature` inside `JSch`'s
static initializer. That is gone; the runtime is identified from `java.vm.name`/`java.vendor` and
nothing is constructed at class initialization time.

## Behaviour on other runtimes

Unchanged. Measured on a JDK 8 JVM (1.8.0_504) against 2.28.7 and against this branch:

| config key | 2.28.7 | this branch |
|---|---|---|
| `ssh-ed25519` | `bc.SignatureEd25519` | `bc.SignatureEd25519` |
| `ssh-ed448` | `bc.SignatureEd448` | `bc.SignatureEd448` |
| `keypairgen.eddsa` | `bc.KeyPairGenEdDSA` | `bc.KeyPairGenEdDSA` |

The only difference is for anyone who deliberately configures the `jce` class on a runtime without
EdDSA: the constructor used to throw `UnsupportedOperationException`, and now `init()` throws
`NoSuchAlgorithmException`. `Session.checkSignatures()` catches `Exception | LinkageError`, so the
algorithm is marked unavailable either way.

## Tests

- `com.jcraft.jsch.jce.SignatureEdDSATest` signs with this implementation and verifies with the
  platform's own `Signature`, and the reverse, for Ed25519 and Ed448, plus the short key guard.
  Surefire runs against `target/classes`, where the `META-INF/versions` entries are inert, so these
  exercise the Java 8 classes, which are the ones Android loads.
- `JSchTest` pins the implementation chosen for a standard runtime and for an Android one.
- `EdDSAAndroidTest` runs on an emulator through `.github/workflows/android.yml` (API 33 and 35).

Please read those emulator legs carefully before treating them as proof that signing works. On both
images `signsAndVerifiesWithAndroidJca` is **skipped**: API 33 has no Ed25519 at all, and on the
API 35 image it is only reachable through the key store provider, which will not sign with an
imported key. What those legs do assert is that the Android configuration is selected and that
unsupported curves fail cleanly in `init()` rather than breaking authentication.

The positive evidence that signing works on Android is a physical device, an Android 16 phone,
where Conscrypt offers both services: `ssh-ed25519` publickey authentication and an interactive
session against OpenSSH 10.5, and the RFC 8032 section 7.1 vector reproduced exactly.

## Note

`verify()` is the same unwrapping as the `bc` and Java 15 implementations. I left it alongside them
rather than extracting a shared helper, to keep the change contained. Happy to factor it out if you
would prefer that.


FireBurn:
Sorry I thought it would be a good idea to run CI on Android too, but wasn't sure how to test - hopefully that's it this time

norrisjeremy (changes requested):
I strongly oppose the introduction of all this Gradle & Android specific machinery into the base JSch project.

My suggestion would be to publish a separate Android specific artifact that provides the Android centric SignatureEdDSA implementation.

src/main/java/com/jcraft/jsch/jce/SignatureEd25519.java
```diff
@@ -28,7 +28,18 @@
 
 public class SignatureEd25519 extends SignatureEdDSA {
 
-  public SignatureEd25519() {
-    throw new UnsupportedOperationException("SignatureEd25519 requires Java15+.");
+  @Override
```
norrisjeremy: Since these are Android specific, would it make more sense to put them into a new package named `com.jcraft.jsch.android`?
That way users on regular Java environments don't mistakenly attempt to use them?

android-test/build.gradle.kts
```diff
@@ -0,0 +1,25 @@
+plugins {
```
norrisjeremy: Is there a reason why this test infrastructure has to use Gradle?
Is there not a way to do this using Maven?

.github/workflows/android.yml
```diff
@@ -0,0 +1,45 @@
...
+    - name: Build Android test
+      run: |
+        ./mvnw -B -V -DskipTests package
+        jar --create --file android-test/jsch-android.jar -C target/classes com
+    - name: Enable KVM
+      run: |
+        echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS="static_node=kvm"' \
+          | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+        sudo udevadm control --reload-rules
+        sudo udevadm trigger --name-match=kvm
```
norrisjeremy: The documentation I see indicates this is only supported on the larger Linux runners, which presumably are not free to use.
Since this project is only using the base free Linux runners, I'd be concerned about how slow this workflow would execute?

See:
- https://github.com/ReactiveCircus/android-emulator-runner#running-hardware-accelerated-emulators-on-linux-runners
- https://github.blog/changelog/2023-02-23-hardware-accelerated-android-virtualization-on-actions-windows-and-linux-larger-hosted-runners/

android-test/build.gradle.kts
```diff
@@ -0,0 +1,25 @@
...
+
+  compileOptions {
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
+  }
+}
+
+dependencies {
+  implementation(files("jsch-android.jar"))
+  androidTestImplementation("androidx.test:runner:1.6.2")
```
norrisjeremy: It appears 1.7.0 is the latest version?
Is there a reason to specifically use the older 1.6.2?

.github/workflows/android.yml
```diff
@@ -0,0 +1,45 @@
...
+      uses: actions/setup-java@b6effb05e454b25005698d916606bdc6ffcbf961 # v5.7.0
+      with:
+        distribution: 'temurin'
+        java-version: '25'
+        check-latest: true
+        cache: 'maven'
+    - name: Set up Gradle
+      uses: gradle/actions/setup-gradle@0723195856401067f7a2779048b490ace7a47d7c # v5.0.0
+      with:
+        gradle-version: '9.5.0'
```
norrisjeremy: The latest gradle release appears to be 9.7.1.
Is there a specific reason to use the older 9.5.0 release?

android-test/build.gradle.kts
norrisjeremy: After doing some research, I'm guessing not.

.github/workflows/android.yml
```diff
@@ -0,0 +1,45 @@
+name: Android CI
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  test:
```
norrisjeremy: Is there a way to split this into two jobs: first a build job, and then a test job that uses the api-level matrix?
That way the build process is only executed once instead of four times.
You can see how I set this up in the maven.yml workflow.

norrisjeremy (commented):
Out of curiosity: does Android provide X25519 / X448 implementations too?

If so, would there be value in providing an Android specific implementation of `com.jcraft.jsch.XDH` so that the `curve25519-sha256` & `curve448-sha512` kex algorithms are usable?

android-test/build.gradle.kts
```diff
@@ -0,0 +1,25 @@
...
+  androidTestImplementation("junit:junit:4.13.2")
```
norrisjeremy: I'm guessing there isn't a JUnit 5+ compatible test runner for Android?

